PR #367, opened 2025-09-14
fix bug that outputted empty maf files

FerriolCalvet:


FerriolCalvet:
functional testing passed!

PR #381, opened 2025-10-12
Handle all python script inputs with click

FerriolCalvet:


modules/local/createpanels/captured/main.nf
```diff
@@ -3,10 +3,7 @@ process CREATECAPTUREDPANELS {
     label 'process_single'
     label 'process_medium_high_memory'
 
-    conda "bioconda::pybedtools=0.9.1--py38he0f268d_0"
-    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-            'https://depot.galaxyproject.org/singularity/pybedtools:0.9.1--py38he0f268d_0' :
-            'biocontainers/pybedtools:0.9.1--py38he0f268d_0' }"
+    container "community.wave.seqera.io/library/bedtools_pybedtools_pandas_pip_pruned:78080da05d53636d"
```
FerriolCalvet, line 6: we should make sure that this container gets "documented" in the containers repo and then we can maybe change it here as well

bin/create_panel4sample.py
```diff
@@ -25,10 +25,14 @@ def create_panel4sample(compact_annot_panel_path, depths_path, panel_name, min_d
...
-    depths_path = sys.argv[2]
-    panel_v = sys.argv[3]
-    min_depth = sys.argv[4]
 
-    create_panel4sample(compact_annot_panel_path, depths_path, panel_v, min_depth)
+@click.command()
+@click.option('--compact-annot-panel-path', required=True, type=click.Path(exists=True), help='Input compact annotation panel file (TSV)')
+@click.option('--depths-path', required=True, type=click.Path(exists=True), help='Input depths file (TSV)')
+@click.option('--panel-name', required=True, type=str, help='Panel name for output files')
+@click.option('--min-depth', required=True, type=int, help='Minimum depth threshold')
```
m-huertasp, line 32: [Minor]
This is super minor, but I think that, if using the argument `type=int` with click, it will only accept int values, so there's no need to transform it here: `min_depth = int(min_depth)`

bin/dNdS_preprocess.py
m-huertasp, line 1: [Minor] Since we’re already looking at this, I noticed the variable `filee` in the `prepare_depths` function — maybe it could be renamed to something clearer.

bin/features_2group_genes.py
m-huertasp, line 1: [Minor] At `load_table_to_json` you output three dictionaries but only one of them is used, maybe the other two can be removed so it is clearer.

bin/filter_cohort.py
m-huertasp, line 1: [Mid] It is not about click but it could be useful to take a deeper look at this script in the future to try and organize the script, with a bit more modularization.

bin/mutations_custom_processing.py
m-huertasp: [Minor] Better to use` sys.exit(0) `than `exit(0)`?

m-huertasp:
I was just taking advantage that I was looking at the scripts to add those comments but they are not about the implementation of click.
The only script that I could find without click is [`check_samplesheet.py`](https://github.com/bbglab/deepCSA/blob/input-with-click/bin/check_samplesheet.py)

m-huertasp:
I've done some more tests, and I think everything is working fine! I have used a small pytest script for this -->
[test_click.py](https://github.com/user-attachments/files/23016362/test_click.py)

bin/filter_cohort.py
FerriolCalvet: agree

bin/mutations_custom_processing.py
FerriolCalvet: updated

PR #386, opened 2025-10-28
flag failing omegas

FerriolCalvet:
Missing:
- [ ] ~Annotate the QCs for all omega types~

I ignore this because the QCs are only relevant for omega, not for omegagloballoc

### DIFF
--- a/bin/annotate_omega_failing.py
+++ b/bin/annotate_omega_failing.py
@@ -1,0 +1,209 @@
+#!/usr/bin/env python
+"""
+Annotate omegas file with flagged gene/sample pairs.
+
+Takes an omegas TSV and one or more compiled flagged-cases TSVs and adds two
+columns to the omegas file: `flagged` (True/False) and `flag_reason` (string or
+empty).
+
+"""
+
+import sys
+import click
+from pathlib import Path
+import pandas as pd
+from typing import List, Tuple, Optional
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+
+
+def load_flagged_tables(paths: List[Path]) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    """Read flagged files and return a normalized DataFrame with possible columns:
+    SAMPLE, GENE, reason_exclusion
+    """
+    pieces = []
+    for p in paths:
+        if not p.exists():
+            print(f"Warning: flagged file {p} does not exist, skipping.")
+            continue        
+        df = pd.read_csv(p, sep="\t", header=0, dtype=str)
+        pieces.append(df)
+    if not pieces:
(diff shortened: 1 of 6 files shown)


workflows/deepcsa.nf
```diff
@@ -560,13 +560,14 @@ workflow DEEPCSA{
                         DNA2PROTEINMAPPING.out.depths_exons_positions.first(),
                         group_keys_ch
                         )
-    }    
-    
-    if ( run_mutdensity ){
+    }
+
+    if ( run_mutdensity && params.omega ){
```
m-huertasp: This could be a little redundant.
We have on one side:
`def run_mutdensity = (params.mutationdensity || params.omega)`

So the condition here really is:
`if ( (params.mutationdensity || params.omega) && params.omega )`

(right?)

Which means that this is going to run if `params.omega` is true regardless of what it says in `params.mutationdensity`.

I understand that this will only work if the inputs are there, which means that `params.mutationdensity` will need to be true, but that is not being checked there, I think.

Total, que it works but could be a bit confusing.

bin/annotate_omega_failing.py
```diff
@@ -0,0 +1,377 @@
...
+            return
+        _heatmap_one(all_df, output_prefix)
+    elif mode == 'separate':
+        _heatmap_one(syn_flagged.copy() if syn_flagged is not None else pd.DataFrame(columns=['sample','gene','reason_exclusion']), f"{output_prefix}_syn")
+        _heatmap_one(npa_flagged.copy() if npa_flagged is not None else pd.DataFrame(columns=['sample','gene','reason_exclusion']), f"{output_prefix}_npa")
+    else:
+        raise ValueError(f"Unknown mode for plot_flagged_heatmap: {mode}")
+
+
+def annotate(omegas: pd.DataFrame, flagged: pd.DataFrame) -> pd.DataFrame:
```
m-huertasp, line 298: I still don't understand why but all those with `reason_exclusion -> mutdensity = 0` are not flagged in the output (`omega.flagged_annotated.tsv`), there are only two flag reasons:

Image

Should it work like this?

m-huertasp (commented):
A lot of work and things here!

I have some questions (below) on how this should work. You'll see it in the comments.
I am not sure if I am missing something but is this working as you expect?

workflows/deepcsa.nf
FerriolCalvet: so we could put params.omega only no?

bin/annotate_omega_failing.py
FerriolCalvet: because if they do not have any synonymous mutation omega is not computed, but very good point I did not check this, but it makes sense

workflows/deepcsa.nf
FerriolCalvet:
```suggestion
    if ( params.omega ){
```

PR #390, opened 2025-11-14
Dev chunk optimization postprocessveppanel

### LINKED ISSUE #114: Parallelize panel annotation steps by splitting the input into different chunks

### LINKED ISSUE #161: Test if it works with the duplexome
This pipeline has been developed with a focus on analyzing data from panels, but it could potentially be applied to any other scenario including exome, or even whole genome although the main interest would be in exonic regions probably.

It would be good that whenever we have the duplexome run finished we can test how deepCSA works with that and fix/optimize any new errors or runtime bottlenecks.

I expect that the annotation of the panel will be very slow, but can be easily parallelized, and many other steps that require the full panel to work with might be tricky to run.

### LINKED ISSUE #262: Adjust resource requirements of different processes to minimize burden for the cluster
- [ ] Define "categories" of processes that require approximately the same amount of resources so that we can dynamically, based on the user configuration (taking into account the panel) adjust the resources
- [ ] Document how the users should choose these amount of resources

### LINKED ISSUE #407: Large memory usage by panel_postprocessing_annotation.py
Performing Duplexome level analysis. The generated captured_panel.tmp.gz file is ~18GB in size.  The process eventually uses up all the available memory  (>500GB) and is killed by the OOM killer (or Slurm, depending). I do get the following warning, which may be relevant to the issue:

`/mnt/home/scottrk/.nextflow/assets/bbglab/deepCSA/bin/panel_postprocessing_annotation.py:143: DtypeWarning: Columns (7) have mixed types. Specify dtype option on import or set low_memory=False.`

migrau:
[copilot generated]

# Performance Optimization: Chunked Processing for Large Panel Annotations

## Overview

This PR introduces memory-efficient chunked processing for VEP annotation post-processing, enabling the pipeline to handle arbitrarily large panel annotations without memory constraints.

#### 1. **`panel_postprocessing_annotation.py` - Chunked VEP Output Processing**

- **Chunk size:** 100,000 lines
- **Implementation:** Streaming pandas read with incremental output writing
- **Benefits:**
  - Processes large VEP outputs without loading entire file into memory
  - Prevents OOM errors on panels with millions of variants
  - Maintains same output quality with predictable resource usage

**Technical details:**
```python
chunk_size = 100000
reader = pd.read_csv(VEP_output_file, sep="\t", chunksize=chunk_size)

for i, chunk in enumerate(reader):
    processed_chunk = process_chunk(chunk, chosen_assembly, using_canonical)
    # Incremental write with header only on first chunk
    rich_out_file.write(processed_chunk.to_csv(header=(i == 0), index=False, sep="\t"))
    del processed_chunk
    gc.collect()  # Explicit memory cleanup
```

**Process:** `CREATEPANELS:POSTPROCESSVEPPANEL`
- Takes per-chromosome output from `VCFANNOTATEPANEL`
- Processes in 100k-line chunks
- **Status:** ✅ Working successfully

#### 2. **`panel_custom_processing.py` - Chromosome-Based Chunked Loading**

- **Chunk size:** 1,000,000 lines
- **Strategy:** Load only relevant chromosome data in chunks
- **Benefits:**
  - Memory-efficient custom region annotation
  - Filters during read to minimize memory footprint

**Technical details:**
```python
def load_chr_data_chunked(filepath, chrom, chunksize=1_000_000):
    reader = pd.read_csv(filepath, sep="\t", chunksize=chunksize, dtype={'CHROM': str})
    chr_data = []
    for chunk in reader:
        filtered = chunk[chunk["CHROM"] == chrom]
        if not filtered.empty:
            chr_data.append(filtered)
    return pd.concat(chr_data) if chr_data else pd.DataFrame()
```

**Process:** `CUSTOMPROCESSING` / `CUSTOMPROCESSINGRICH`
- Processes custom genomic regions with updated annotations
- Loads data per-chromosome to reduce memory usage

### ❌ VEP Cache Storage Location - No Performance Impact

**What was tested:**
- Using VEP cache from beegfs storage (`/workspace/datasets/vep` or `/data/bbg/datasets/vep`)
- Expected faster cache access vs. downloading on-the-fly

**Results:**
- **No significant runtime improvement** for `ENSEMBLVEP_VEP` process
- VEP annotation runtime is compute-bound, not I/O-bound
- Network-attached storage performed equivalently to local cache
- OS filesystem caching likely mitigates storage location differences

**Commits:**
- `035a0c7` (April 3, 2025): Added VEP cache beegfs support
- `8e40d83` (April 24, 2025): Removed VEP cache beegfs optimization (no benefit)

**Current approach:** 
- Cache location configurable via `params.vep_cache`
- Defaults to downloading cache if not provided
- Various config files specify beegfs paths for convenience, not performance

## Resource Configuration

Updated resource limits for chunked processes:

```nextflow
withName: '(BBGTOOLS:DEEPCSA:CREATEPANELS:POSTPROCESSVEPPANEL*|...)' {
    cpus   = { 2 * task.attempt }
    memory = { 4.GB * task.attempt }
    time   = { 360.min * task.attempt }
}
```

## Integration Points

**Affected Subworkflows:**
- `CREATEPANELS` → `POSTPROCESSVEPPANEL` → processes VEP output in chunks
- `CUSTOMPROCESSING` / `CUSTOMPROCESSINGRICH` → uses chunked loading for custom regions

**Pipeline Flow:**
```
SITESFROMPOSITIONS → VCFANNOTATEPANEL (VEP) 
    ↓
POSTPROCESSVEPPANEL (chunked processing) ← 100k line chunks
    ↓
CUSTOMPROCESSING (optional, chunked by chromosome)
    ↓
CREATECAPTUREDPANELS / CREATESAMPLEPANELS / CREATECONSENSUSPANELS
```

## Testing

**Tested on:**
- Large-scale panels (millions of variants)
- Multiple configuration profiles (nanoseq, chip, kidney, etc.)

**Validation:**
- Output correctness verified (same results as non-chunked version)
- Memory usage remains stable across panel sizes
- No OOM errors on large inputs

## Performance Impact

| Metric | Before | After |
|--------|--------|-------|
| **Memory usage** | Unbounded (full file in RAM) | ~4 GB (controlled) |
| **Max panel size** | Limited by available memory | Unlimited |
| **Runtime** | Similar | Similar (no regression) |
| **Reliability** | OOM on large panels | Stable processing |

## Migration Notes

No breaking changes. Existing pipelines continue to work with improved memory efficiency.

## Related Commits

- `276152d`: Chunking for `panel_custom_processing.py`
- `035a0c7`: VEP cache beegfs attempt (added)
- `8e40d83`: VEP cache beegfs removal (no performance gain)
- Various fixes: `1dffd94`, `945c129`, `d243ebc`, etc. (resource tuning)

## Conclusion

This PR successfully implements memory-efficient chunked processing for panel annotation post-processing, enabling the pipeline to scale to arbitrarily large panels without memory constraints. The VEP cache storage location experiment confirmed that computation, not I/O, is the bottleneck for annotation runtime.


bin/create_consensus_panel.py
FerriolCalvet, line 1: these changes may not be required since I already updated the Nextflow module to make the failing consensus file optional.

I think I would prefer to not generate the file if there is nothing to report.

conf/base.config
FerriolCalvet, line 1: this update looks good, but I am curious to know based on which samples was this defined.
It would be great to make sure that it works well for the last 2 duplexomes and the kidney cohort with the pancancer panel for example

conf/base.config
```diff
@@ -1,27 +1,28 @@
...
+  // === RESOURCE LIMITS ===
+    resourceLimits = [ 
+        cpus: params.max_cpus, 
+        memory: params.max_memory, 
+        time: params.max_time 
+    ]
+
+    // === SENSIBLE DEFAULTS ===
+    // Most processes use minimal resources based on usage analysis
+    cpus   = { 1 }
```
FerriolCalvet, line 11: I think this is OK, but I think we should revise that all the steps that might be able to use multiple threads get at least the chance of increasing the number of CPUs in the new attempts

(nothing to change just a heads-up on this topic)

conf/nanoseq.config
FerriolCalvet, line 1: when is this one used?

bin/panel_custom_processing.py
```diff
@@ -88,25 +119,25 @@ def customize_panel_regions(VEP_output_file, custom_regions_file, customized_out
...
-    all_possible_sites.to_csv(customized_output_annotation_file,
-                                        header = True,
-                                        index = False,
-                                        sep = "\t")
+    chr_data = chr_data.drop_duplicates(
+        subset=['CHROM', 'POS', 'REF', 'ALT', 'MUT_ID', 'GENE', 'CONTEXT_MUT', 'CONTEXT', 'IMPACT'],
+        keep='first'
+    )
+    chr_data.to_csv(customized_output_annotation_file, header=True, index=False, sep="\t")
+
```
FerriolCalvet: I am not sure this does the same as it was doing before, because it is supposed to output all the same TSV table but replacing the values in some of the rows, in this case it seems that only the information from the last chromosome will be outputted, but maybe I got it wrong

modules/local/process_annotation/panelcustom/main.nf
```diff
@@ -30,10 +27,16 @@ process CUSTOM_ANNOTATION_PROCESSING {
     // neutral_impact      : (optional, default; synonymous)
     // new_impact          : (optional, default: missense) is the impact that the mutations listed in impactful_mutations will receive.
     """
+    # Calculate expected number of chunks
+    n_lines=\$(wc -l < ${panel_annotated})
+    n_chunks=\$(( (n_lines + ${chr_chunk_size} - 1) / ${chr_chunk_size} ))
+    echo "[CUSTOM_ANNOTATION_PROCESSING] Processing ${meta.id} with internal chr_chunk_size=${chr_chunk_size} (\${n_lines} lines, ~\${n_chunks} chunks)"
+
```
FerriolCalvet: I am not sure that these lines are needed but could be helpful for debugging

modules/local/sitesfrompositions/main.nf
```diff
@@ -34,7 +31,32 @@ process SITESFROMPOSITIONS {
...
+        
+        n_chunks=\$(ls captured_positions.sites4VEP.chunk*.tsv | wc -l)
+        echo "[SITESFROMPOSITIONS] Created \${n_chunks} chunks"
+        
+        rm header.tmp captured_positions.sites4VEP.full.tsv
+    else
+        echo "[SITESFROMPOSITIONS] No chunking (chunk_size=0), processing as single file"
+        mv captured_positions.sites4VEP.full.tsv captured_positions.sites4VEP.chunk1.tsv
+    fi
+
```
FerriolCalvet: Looks good, but maybe it would be cleaner to implement the chunking in the python script that generates the file with the sites? I don't know what do you think?

subworkflows/local/createpanels/main.nf
```diff
@@ -65,24 +73,44 @@ workflow CREATE_PANELS {
...
+        .map{ file -> [[ id : "captured_panel"], file] }
+        .set{ merged_rich_unsorted }
+
+    // Sort merged panels to ensure genomic order
+    SORTPANELCOMPACT(merged_compact_unsorted)
+    SORTPANELRICH(merged_rich_unsorted)
+
+    merged_compact = SORTPANELCOMPACT.out.sorted
+    merged_rich = SORTPANELRICH.out.sorted
+
```
FerriolCalvet, line 83: these looks good, but now you should add these two new variables to the outputs listed below, since otherwise only a chunk of the panels will be outputted, see lines 189 and 190 below

tests/deepcsa.nf.test
```diff
@@ -58,10 +58,13 @@ nextflow_pipeline {
...
-            assert header.contains("gene") : "Omega output should contain 'gene' column"
-            assert header.contains("sample") : "Omega output should contain 'sample' column"
-            assert header.contains("dnds") : "Omega output should contain 'dnds' column"
+            // Skip empty lines at the beginning (can happen with collectFile)
+            // def headerLine = lines.find { it.trim() != "" }
+            // assert headerLine != null : "Omega output should contain a header"
+            // def header = headerLine.split('\t')
+            // assert header.contains("gene") : "Omega output should contain 'gene' column"
+            // assert header.contains("sample") : "Omega output should contain 'sample' column"
+            // assert header.contains("dnds") : "Omega output should contain 'dnds' column"
```
FerriolCalvet: with the update in omega, we could check for a snapshot of the file here as well

tests/nextflow.config
```diff
@@ -33,6 +33,9 @@ executor {
 }
 
 params {
+    panel_postprocessing_chunk_size          = 100000000
```
FerriolCalvet: I would remove this parameter since it is complex to manage it properly

```suggestion
```

nextflow.config
```diff
@@ -104,6 +104,9 @@ params {
     min_muts_per_sample                      = 0
     selected_genes                           = ''
     panel_with_canonical                     = true
+    panel_postprocessing_chunk_size          = 100000 // a very big number will avoid chunking by default
```
FerriolCalvet: as I said in other places, I would remove this parameter

```suggestion
```

nextflow.config
```diff
@@ -185,9 +188,9 @@ params {
 
     // Max resource options
     // Defaults only, expecting to be overwritten
-    max_memory                  = 256.GB
-    max_cpus                    = 56
-    max_time                    = 240.h
+    max_memory                  = 950.GB
+    max_cpus                    = 196
+    max_time                    = 30.d
```
FerriolCalvet, line 190: I understand this needs to be changed by the user, but maybe we should switch it to more realistic thresholds no?

nextflow_schema.json
```diff
@@ -559,6 +559,35 @@
...
+                    "fa_icon": "fas fa-cut",
+                    "help_text": "When set to a positive integer, splits the sites file into chunks for parallel processing through VEP annotation. Set to 0 to disable chunking (process as single file). Recommended values: 100000-500000 for large datasets."
+                },
+                "panel_postprocessing_chunk_size": {
+                    "type": "integer",
+                    "description": "Internal chunk size for VEP postprocessing memory management",
+                    "default": 100000,
+                    "fa_icon": "fas fa-memory",
+                    "help_text": "Controls how the panel_postprocessing_annotation.py script processes data internally. Higher values use more memory but may be faster. Not related to file-level chunking."
+                },
```
FerriolCalvet:
```suggestion
```

FerriolCalvet (changes requested):
I went over all the files and these are some of the comments, in general I think that these are the main points:

- [x] one bigger change is to not parallelize the processing of Ensembl VEP annotation, but keep the paralellization to splitting the input.
- [x] Also the chunking for the custom processing of the panel is a good idea but I am not sure that the implementation is correct, it should be revised.
- [x] Add omega snapshot as part of the test

Once these details are solved, it would be great to merge the dev branch here (solve conflicts) and confirm that all the tests are passing
- [ ] Merge with the dev branch and update the tests snapshots in case it is needed

subworkflows/local/createpanels/main.nf
migrau: Fixed

modules/local/sitesfrompositions/main.nf
migrau: Makes sense, done!

conf/nanoseq.config
migrau: An an optional mode in the nextflow.config, yes. But I will re-do some tests to check if we can use only the base one, and discard this one.

conf/base.config
migrau: I will repeat the tests with duplexome and re-adjust if required.

conf/base.config
migrau: From a quick check, pending to review: OMEGA_ESTIMATOR and SIGPROFILERASSIGNMENT

bin/panel_custom_processing.py
migrau: true. Reviewed and improved.

bin/create_consensus_panel.py
migrau: Perfect. `else` removed

tests/deepcsa.nf.test
migrau: added

m-huertasp:
Hi! While checking the cord bloods run (combining DupCaller and deepUMI callings) I saw that one of the places in which we have a bottleneck is in the `postprocessveppanel` and I found this PR. I've checked the python script and I think it could also be a bit more optimized by using `polars` instead of `pandas` (now that we have it in the deepCSA container) and trying to change the "apply" logic in some places. Just a "heads-up" to say that it could be done in the future so I don't forget.

nextflow_schema.json
```diff
@@ -310,6 +310,11 @@
                     "description": "Do you want to plot the depths within each gene?",
                     "fa_icon": "fas fa-book"
                 },
+                "plot_mutation_specific_qc": {
+                    "type": "boolean",
+                    "description": "Do you want to generate mutation-specific QC plots (VAF vs depth)?",
+                    "fa_icon": "fas fa-book"
```
FerriolCalvet:
```suggestion
                    "fa_icon": "fas fa-book",
                    "hidden": true,
                    "default": true
```

FerriolCalvet: I would leave this as true by default and hide it

nextflow_schema.json
```diff
@@ -583,6 +588,28 @@
...
+        "parallel_processing_parameters": {
+            "title": "Parallel processing and chunking options",
+            "type": "object",
+            "fa_icon": "fas fa-tasks",
+            "description": "Parameters to control parallel processing, chunking, and memory management during panel creation and annotation.",
+            "properties": {
+                "panel_sites_chunk_size": {
+                    "type": "integer",
+                    "description": "Number of sites per chunk for parallel VEP annotation (0 = no chunking)",
+                    "default": 0,
```
FerriolCalvet:
```suggestion
                    "default": 1000000,
```

nextflow.config
```diff
@@ -105,6 +106,8 @@ params {
     min_muts_per_sample                      = 0
     selected_genes                           = ''
     panel_with_canonical                     = true
+    panel_custom_processing_chunk_size       = 1000000 // a very big number will avoid chunking by default
+    panel_sites_chunk_size                   = 0  // 0 means no chunking (default), set to positive integer to enable chunking
```
FerriolCalvet:
```suggestion
    panel_sites_chunk_size                   = 1000000  // 0 means no chunking (default), set to positive integer to enable chunking
```

subworkflows/local/createpanels/main.nf
```diff
@@ -50,25 +58,45 @@ workflow CREATE_PANELS {
...
+        .collectFile(name: 'captured_panel.vep.annotation.tsv', keepHeader: true, skip: 1)
+        .map{ file -> [[ id : "captured_panel"], file] }
+        .set{ merged_compact_unsorted }
+
+    POSTPROCESSVEPPANEL.out.rich_panel_annotation
+        .map{ it[1] }
+        .collectFile(name: 'captured_panel.vep.annotation.rich.tsv', keepHeader: true, skip: 1)
+        .map{ file -> [[ id : "captured_panel"], file] }
+        .set{ merged_rich_unsorted }
+
```
FerriolCalvet, line 76:
```suggestion
    POSTPROCESSVEPPANEL.out.compact_panel_annotation
        .map{ it -> it[1] }
        .collectFile(name: 'captured_panel.vep.annotation.tsv', keepHeader: true, skip: 1)
        .map{ file -> [[ id : "captured_panel"], file] }
        .set{ merged_compact_unsorted }

    POSTPROCESSVEPPANEL.out.rich_panel_annotation
        .map{ it -> it[1] }
        .collectFile(name: 'captured_panel.vep.annotation.rich.tsv', keepHeader: true, skip: 1)
        .map{ file -> [[ id : "captured_panel"], file] }
        .set{ merged_rich_unsorted }

```

subworkflows/local/createpanels/main.nf
```diff
@@ -38,10 +40,16 @@ workflow CREATE_PANELS {
     // Create all possible sites and mutations per site of the captured panel
     SITESFROMPOSITIONS(depths)
 
-    // Create a tuple for VEP annotation (mandatory)
-    SITESFROMPOSITIONS.out.annotated_panel_reg.map{ it -> [[ id : "captured_panel"],  it[1]] }.set{ sites_annotation }
+    // Flatten chunks and create tuples for VEP annotation
+    SITESFROMPOSITIONS.out.annotated_panel_reg
+        .transpose()
+        .map{ meta, chunk ->
```
FerriolCalvet:
```suggestion
        .map{ _meta, chunk -> 
```

subworkflows/local/createpanels/main.nf
```diff
@@ -50,25 +58,45 @@ workflow CREATE_PANELS {
...
 
         added_regions = CUSTOMPROCESSINGRICH.out.added_regions
 
     } else {
-        complete_annotated_panel = POSTPROCESSVEPPANEL.out.compact_panel_annotation
-        rich_annotated = POSTPROCESSVEPPANEL.out.rich_panel_annotation
-        added_regions = channel.empty()
+        complete_annotated_panel = merged_compact
+        rich_annotated = merged_rich
+        added_regions = Channel.empty()
```
FerriolCalvet:
```suggestion
        added_regions = channel.empty()
```

modules/local/createpanels/consensus/main.nf
```diff
@@ -1,6 +1,5 @@
 process CREATECONSENSUSPANELS {
     tag "$meta.id"
-    label 'process_single'
 
     conda "python=3.10.17 bioconda::pybedtools=0.12.0 conda-forge::polars=1.30.0 conda-forge::click=8.2.1 conda-forge::gcc_linux-64=15.1.0 conda-forge::gxx_linux-64=15.1.0"
     container 'docker://bbglab/deepcsa_bed:latest'
```
FerriolCalvet, line 5: I think that the receipe of this container is not pushed to https://github.com/bbglab/containers-recipes
if you have it localized somewhere try to push it so that we have everything centralized there, but go ahead with the merge

FerriolCalvet (commented):
Looks good Miguel!

I left some comments and suggestions.
nothing critical.
* only some minor fixes to pass the nextflow linting
* update default chunk_size to 1M so that bigger panels get chunked

and then another comment is that maybe we might need to be more generous in terms of memory of some steps when running bigger cohorts, but we will see this as we start using it.

I would apply the suggestions if you agree and then merge it to dev so that it starts to get tested by all of us and we tune it from there

thankss!!

FerriolCalvet:
Wait Miguel, we found some weird behaviour in the test run with the cord bloods I will let you know once we solve it

bin/panel_custom_processing.py
FerriolCalvet: this is the error that is still not fixed, only the information from one chromosome is outputted.

I will work now on solving it

FerriolCalvet: @migrau I see this was not fixed the easiest way to handle it for me was to revert the change to the previous implementation here : https://github.com/bbglab/deepCSA/tree/custom-processing-fix

migrau:
the results from omega were not deterministic, e.g:
image

Now, the test rounds numeric columns (dnds, pvalue, lower, upper) to 2 decimals before comparison.

On the other hand, the implementation of polars in `bin/panel_custom_processing.py` is much faster than pandas but it requires ~30% of RAM, because there is not chunking. Results tested and compared with the previous implementation and they have the same md5sum.

migrau:
MAF tests from current dev added  https://github.com/bbglab/deepCSA/pull/390/commits/330690e1dd9ca6fb12dac85f231b8a079a8dd767

Test passed. New snapshots match the ones from dev. 

```
$ nf-test test 

🚀 nf-test 0.9.2
https://www.nf-test.com
(c) 2021 - 2024 Lukas Forer and Sebastian Schoenherr


Test DEEPCSA Pipeline

  Test [4150b7be] 'Minimal features test run' PASSED (677.383s)
  Test [9a4541aa] 'MAF-based minimal features test run' PASSED (1019.239s)
  Test [1e61cc82] 'Omega analysis test run' PASSED (651.906s)
  Test [c650506f] 'MAF-based omega analysis test run' PASSED (1006.037s)
  Test [b4f95eff] 'Mutation density test run' PASSED (652.338s)
  Test [89f4b5ca] 'MAF input validation - fails when --input_maf is provided without --use_custom_depths' PASSED (7.543s)


SUCCESS: Executed 6 tests in 4014.461s

(nf-test) [deepCSA] [mgrau@irbccn34 deepCSA]
```

migrau:
merge FROM dev done and test passed:

```
$ nf-test test 

🚀 nf-test 0.9.2
https://www.nf-test.com
(c) 2021 - 2024 Lukas Forer and Sebastian Schoenherr

Load .nf-test/plugins/nft-utils/0.0.3/nft-utils-0.0.3.jar

Test DEEPCSA Pipeline

  Test [4150b7be] 'Minimal features test run' PASSED (625.354s)
  Test [9a4541aa] 'MAF-based minimal features test run' PASSED (1050.716s)
  Test [1e61cc82] 'Omega analysis test run' PASSED (665.905s)
  Test [c650506f] 'MAF-based omega analysis test run' PASSED (1057.473s)
  Test [b4f95eff] 'Mutation density test run' PASSED (619.715s)
  Test [89f4b5ca] 'MAF input validation - fails when --input_maf is provided without --use_custom_depths' PASSED (8.228s)


SUCCESS: Executed 6 tests in 4036.851s
```

We can go ahead with the merge to dev

PR #401, opened 2025-12-14
Add mutdensity subgenic and fix bugs

### LINKED ISSUE #400: If DNA2PROTEINMAPPING fails, blocks execution of omega
Whenever DNA2PROTEINMAPPING exits with an error, this prevents the execution of the EXPANDREGIONS step and then this stops the rest...

this should not be the case, so let's fix it.

FerriolCalvet:
This PR includes the restructurization of the calls to EXPANDREGIONS to obtain a value of mutation density for subgenic regions defined from exons or domains.
It also includes some bug fixing for cases where the DNA2PROTEINMAPPING fails and then omega was not executed leading to other errors downstream.

## AI summary
This pull request introduces a new workflow for enriching genomic panels with subgenic regions and refactors how panels are expanded and used throughout the pipeline. It also upgrades the `omega` tool version and improves reproducibility and flexibility in several processes. The changes streamline the workflow, making panel enrichment modular and easier to maintain, and ensure that downstream analyses use the correct, enriched panels.

**Workflow and Pipeline Refactoring**

* Added a new `ENRICHPANELS` workflow (`subworkflows/local/enrichpanels/main.nf`) to handle the enrichment of consensus panels with subgenic regions, domains, and exons. This workflow centralizes and modularizes region expansion logic.
* Refactored the main workflow (`workflows/deepcsa.nf`) to use the new `ENRICHPANELS` workflow for generating expanded panels and subgenic region JSONs, and updated downstream processes (e.g., mutation density, OMEGA analysis) to use these enriched panels instead of the original ones. [[1]](diffhunk://#diff-87c636ed11c630b436dd28dc585de64022c0bbf406a7875587e2598a5a0c9cb5R26-R28) [[2]](diffhunk://#diff-87c636ed11c630b436dd28dc585de64022c0bbf406a7875587e2598a5a0c9cb5L231-R247) [[3]](diffhunk://#diff-87c636ed11c630b436dd28dc585de64022c0bbf406a7875587e2598a5a0c9cb5L255-R267) [[4]](diffhunk://#diff-87c636ed11c630b436dd28dc585de64022c0bbf406a7875587e2598a5a0c9cb5L406-R421)
* Removed the direct use of the `EXPAND_REGIONS` process from the OMEGA analysis subworkflow, delegating all panel expansion to the new enrichment workflow. [[1]](diffhunk://#diff-05ea1fa601c398963ce647cd3e32208eaf090bbd14563fee67cb35108ca8308aL8) [[2]](diffhunk://#diff-05ea1fa601c398963ce647cd3e32208eaf090bbd14563fee67cb35108ca8308aL33-L52) [[3]](diffhunk://#diff-05ea1fa601c398963ce647cd3e32208eaf090bbd14563fee67cb35108ca8308aL75-L83)

**Process and Container Updates**

* Updated all OMEGA-related processes (`OMEGA_PREPROCESS`, `OMEGA_MUTABILITIES`, `OMEGA_ESTIMATOR`) to use the new `docker.io/bbglab/omega:0.2.1` container instead of the previous one, and improved version reporting by dynamically capturing the actual tool version at runtime. [[1]](diffhunk://#diff-ed3dff2bfca28aeedcd7ce6b22127b7427f14a7f581e338285f83c7cdd04ab46L3-R8) [[2]](diffhunk://#diff-8a1335560d13972beb8017e9984dfdb006d840a0552f1fdb0a91a2f2fb867d6dL8-R8) [[3]](diffhunk://#diff-28820cc8bc53dc1436d38748b7d21e053765d47ec690727cfe87357d192dfe95L8-R8) [[4]](diffhunk://#diff-ed3dff2bfca28aeedcd7ce6b22127b7427f14a7f581e338285f83c7cdd04ab46L56-R70) [[5]](diffhunk://#diff-8a1335560d13972beb8017e9984dfdb006d840a0552f1fdb0a91a2f2fb867d6dL40-R43) [[6]](diffhunk://#diff-8a1335560d13972beb8017e9984dfdb006d840a0552f1fdb0a91a2f2fb867d6dL55-R55) [[7]](diffhunk://#diff-28820cc8bc53dc1436d38748b7d21e053765d47ec690727cfe87357d192dfe95L44-R44) [[8]](diffhunk://#diff-28820cc8bc53dc1436d38748b7d21e053765d47ec690727cfe87357d192dfe95L57-L73)

**Panel Expansion and Subgenic Regions**

* Replaced all references to "hotspots" with "subgenic" regions in the `EXPAND_REGIONS` process, including output file names and script logic, to better reflect the expanded panel content. [[1]](diffhunk://#diff-f092a217bbdc631c9b8fc6a57da89ded663e341104bb0f80678faab3ab989227L15-R26) [[2]](diffhunk://#diff-f092a217bbdc631c9b8fc6a57da89ded663e341104bb0f80678faab3ab989227L41-R41)
* Improved logic for handling optional inclusion of exons, domains, and custom subgenic regions, making the workflow more flexible and robust to different parameter settings. [[1]](diffhunk://#diff-5b03bcbf6d20655d43252df969d2ecd6a9d899cbbc310da9735a509dfd8a3700R1-R92) [[2]](diffhunk://#diff-f092a217bbdc631c9b8fc6a57da89ded663e341104bb0f80678faab3ab989227L15-R26)

**Error Handling and Configuration**

* Added a custom error strategy for the `DNA2PROTEINMAPPING` process, allowing up to two retries before ignoring errors, which increases pipeline robustness.

**Miscellaneous Improvements**

* Simplified and clarified output handling in various subworkflows, ensuring that only the first file from multi-file outputs is used where appropriate.
* Cleaned up and removed unused or outdated code and comments for better maintainability. [[1]](diffhunk://#diff-28820cc8bc53dc1436d38748b7d21e053765d47ec690727cfe87357d192dfe95L57-L73) [[2]](diffhunk://#diff-ed3dff2bfca28aeedcd7ce6b22127b7427f14a7f581e338285f83c7cdd04ab46L35-L38)

These changes collectively improve the modularity, maintainability, and reliability of the pipeline, especially around the management and enrichment of gene panels.


modules/local/expand_regions/main.nf
```diff
@@ -12,19 +12,18 @@ process EXPAND_REGIONS {
...
-    path "versions.yml"                         , topic: versions
+    tuple val(meta), path("*with_subgenic.tsv")  , emit: panel_increased
+    tuple val(meta), path("subgenic_names.json") , emit: new_regions_json
+    path "versions.yml"                          , topic: versions
 
 
     script:
-    // def expansion = task.ext.expansion ?: 0
-    def autoexons = params.omega_autoexons ? "--autoexons ${exons}" : ""
+    def autoexons = (params.omega_autoexons & exons.size() > 0 ) ? "--autoexons ${exons}" : ""
```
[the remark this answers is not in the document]
m-huertasp: In fact, I think that this is causing a FAILURE when doing the "Should run with autoexons and autodomains" test in the tests for the [expand_regions](https://github.com/bbglab/deepCSA/tree/main/modules/local/expand_regions) module.

I think that it is a good idea to generalize this tests in the pipeline, by the way.

subworkflows/local/createpanels/main.nf
m-huertasp, line 86: [Minor] I get a warning here:

```
Warn  subworkflows/local/createpanels/main.nf:86:25: The use of `Channel` to access channel factories is deprecated -- use `channel` instead
│  86 |         added_regions = Channel.empty()
╰     | 
```

subworkflows/local/omega/main.nf
```diff
@@ -30,26 +29,17 @@ workflow OMEGA_ANALYSIS{
...
-    domains_ch = params.omega_autodomains ? domains_file : []  // .map{ it -> it[1]} : []
-    exons_ch = params.omega_autoexons ? exons_file.map{ it -> it[1]} : []
-
-    // Create a channel for the hotspots bedfile if provided
-    subgenic_ch = params.omega_subgenic_bedfile ? file(params.omega_subgenic_bedfile) : []
-
-
     site_comparison_results = Channel.empty()
     global_loc_results      = Channel.empty()
     all_gloc_results        = Channel.empty()
```
m-huertasp, line 45: Same warning here:

```
Warn  subworkflows/local/omega/main.nf:43:31: The use of `Channel` to access channel factories is deprecated -- use `channel` instead
│  43 |     site_comparison_results = Channel.empty()
╰     |                               ^^^^^^^

Warn  subworkflows/local/omega/main.nf:44:31: The use of `Channel` to access channel factories is deprecated -- use `channel` instead
│  44 |     global_loc_results      = Channel.empty()
╰     |                               ^^^^^^^

Warn  subworkflows/local/omega/main.nf:45:31: The use of `Channel` to access channel factories is deprecated -- use `channel` instead
│  45 |     all_gloc_results        = Channel.empty()
╰     |                               ^^^^^^^

```

subworkflows/local/omega/main.nf
m-huertasp, line 172: I know this is not included in this PR, but I get this warning while checking:

```
Warn  subworkflows/local/omega/main.nf:171:20: Implicit closure parameter is deprecated, declare an explicit parameter instead
│ 171 |         def meta = it[0]
╰     |                    ^^

Warn  subworkflows/local/omega/main.nf:172:25: Implicit closure parameter is deprecated, declare an explicit parameter instead
│ 172 |         def all_files = it[1..-1].flatten()
```

bin/add_subgenicregions.py
```diff
@@ -152,11 +152,12 @@ def main(panel_file, autoexons, autodomains, custom, subgenic_regions_complement
...
-
-    final_data.to_csv("exons_consensus_panel_with_hotspots.tsv",
+    
+    panel_name = ".".join(panel_file.split(".")[:-1])
+    final_data.to_csv(f"{panel_name}_with_subgenic.tsv",
                         sep="\t", header=True, index=False)
 
-    with open("hotspot_names.json", "w") as f:
+    with open("subgenic_names.json", "w") as f:
         json.dump({x: [x] for x in hotspots_names}, f, indent=4)
```
m-huertasp: In this file there are some variables that are still named with "hotspots", maybe we could rename all of them to avoid confusion?

bin/add_hotspots.py
m-huertasp, line 140: I know this is outside the scope of this PR, but a quick and easy efficiency improvement would be to collect the `complement_rows` into a list and concatenate them all at once.

bin/compute_mutdensity.py
```diff
@@ -143,8 +145,17 @@ def load_n_process_inputs(maf_path, depths_path, annot_panel_path, sample_name):
     depths_subset_adj_sample_df = depths_df.merge(annot_panel_df.drop_duplicates(subset = ["CHROM", "POS", "REF", "ALT"])[["CHROM", "POS"]],
                                                     on = ["CHROM", "POS"], how = "inner")
 
-    return maf_df, depths_subset_df, depths_subset_adj_df, depths_subset_adj_sample_df
+    # Add domains and exons to maf_df
+    annot_panel_df['CHROM_POS'] = [f'{chrom}:{pos}' for chrom, pos in zip(annot_panel_df['CHROM'], annot_panel_df['POS'])]
+    maf_df_raw['CHROM_POS'] = [mutid.split('_')[0] for mutid in maf_df_raw['MUT_ID']]
```
m-huertasp: I understand that there is only one mutid in each row, is this right? In this case, maybe we could do:

```suggestion
    maf_df_raw['CHROM_POS'] = maf_df_raw['MUT_ID'].str.split('_')[0]
```

bin/compute_mutdensity.py
```diff
@@ -143,8 +145,17 @@ def load_n_process_inputs(maf_path, depths_path, annot_panel_path, sample_name):
     depths_subset_adj_sample_df = depths_df.merge(annot_panel_df.drop_duplicates(subset = ["CHROM", "POS", "REF", "ALT"])[["CHROM", "POS"]],
                                                     on = ["CHROM", "POS"], how = "inner")
 
-    return maf_df, depths_subset_df, depths_subset_adj_df, depths_subset_adj_sample_df
+    # Add domains and exons to maf_df
+    annot_panel_df['CHROM_POS'] = [f'{chrom}:{pos}' for chrom, pos in zip(annot_panel_df['CHROM'], annot_panel_df['POS'])]
```
m-huertasp: [Updated] Maybe we could do something like:

```suggestion
    annot_panel_df['CHROM_POS'] = annot_panel_df['CHROM'].astype(str) + ':' + annot_panel_df['POS'].astype(str)
```

The `.astype(str)` is just in case to avoid problems.

bin/omega_select_mutdensity.py
```diff
@@ -14,20 +14,14 @@ def select_syn_mutdensity(mutdensity_file, output_file, mode):
     mutdensity_df = pd.read_csv(mutdensity_file, sep = "\t", header = 0, na_values = custom_na_values)
 
     synonymous_mutdensities_all_samples = mutdensity_df[(mutdensity_df["MUTTYPES"] == "SNV") &
-                                                    (mutdensity_df["GENE"] != "ALL_GENES")].reset_index(drop = True)
+                                                        (mutdensity_df["GENE"] != "ALL_GENES") &
+                                                        ~(mutdensity_df["GENE"].str.contains("--"))].reset_index(drop = True)
```
m-huertasp, line 18: Why are you doing this? What are you filtering out?

m-huertasp (commented):
I think this is quite good job! I might be missing some context regarding where and why we are doing this but I added some comments. If you need any further help, feel free to let me know. I’d also maybe need you to explain the PR in more detail.😃

bin/omega_select_mutdensity.py
FerriolCalvet: all the subgenic regions contain this -- in their names, this is a way of filtering it out

bin/compute_mutdensity.py
FerriolCalvet: uhh I did not know that we could use format in this way, I will apply this suggestion

subworkflows/local/createpanels/main.nf
FerriolCalvet: I am not sure if applying this change will work in 25.04 or only in 25.10 in any case happy to make this change I was aware but intentionally postponing it

FerriolCalvet: I will open an issue and branch dev to apply all the nextflow formatting changes after merging this PR does it make sense to you @m-huertasp

m-huertasp: Totally true, I think they are not going to deprecate this for now so we still have some time (not 100% sure though).

bin/compute_mutdensity.py
FedericaBrando: Love this suggestion! Vector operations on dataframe make the code run faster!

FedericaBrando (approved):
Given Marta's review I did a quick look on the code and it seems all correct. Running a linter every now and then helps with the deprecation errors - also, you can select nextflow version in the IDE to match the error and be sure that it doesn't break in prod. 

Overall keep up the good work! 👊🏻

modules/local/expand_regions/main.nf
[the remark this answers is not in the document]
FerriolCalvet: I made the change but I did not test the test, we should also set up unit tests ASAP for most of the modules...

bin/compute_mutdensity.py
FerriolCalvet: I applied it but it failed and then I had to change it to this:

`maf_df_raw['CHROM_POS'] = maf_df_raw['MUT_ID'].str.split('_').str[0]`

otherwise it was failing because the [0] only took a single item

FedericaBrando: str.split takes an expand argument, then when it set to true it will split the string in columns, then you pick the first one

https://pandas.pydata.org/docs/reference/api/pandas.Series.str.split.html

FerriolCalvet: done!

PR #153, opened 2024-10-23
FEAT: Mask sites where filters are applied

FedericaBrando:
This PR includes several changes to mask the position where we applied these filters (flagged region):

- `repetitive_variant`
- `cohort_n_rich`, `cohort_n_rich_uni`
- `other_samples_snp`

These regions needs to be masked in the depth (depth=0) so that they are not passed to the analysis step.

```zsh
❯ cat all_samples_discarded.bed | head
chr1    11107661        11107661        FILTER.cohort_n_rich_uni
chr1    11108009        11108009        FILTER.cohort_n_rich_uni
chr1    11108412        11108412        FILTER.cohort_n_rich_uni
chr1    11109993        11109993        FILTER.cohort_n_rich_uni,FILTER.other_sample_SNP
chr1    11110011        11110011        FILTER.cohort_n_rich_uni
chr1    11110074        11110074        FILTER.cohort_n_rich_uni,FILTER.other_sample_SNP
```

The most important changes include splitting the `FILTERBATCH` process for better output management, including outputting flagged_regions.bed, adding new input to `ANNOTATE_DEPTHS`, and updating the `DEEPCSA` workflow to manage the dependancies between `MUTATION_PREPROCESSING`  and `ANNOTATE_DEPTHS`.

Related to #53 #118 #117

Closes #118

-----------------------------------------------------------------------------------------------
<details><summary><h3>CHANGES</h3></summary>
<p>

### Changes to `FILTERBATCH` Process:
* Split the `FILTERBATCH` process into its own section to manage its output directory separately (`conf/modules.config`).
* Added a new output path `flagged_muts` to the `FILTER_BATCH` process (`modules/local/filtermaf/main.nf`).

### Changes to `MUTATION_PREPROCESSING` Workflow:
* Emitted `flagged_bed` from the `FILTERBATCH` process in the `MUTATION_PREPROCESSING` workflow (`subworkflows/local/mutationpreprocessing/main.nf`).

### Changes to `DEEPCSA` Workflow:
[...]

PR #412, opened 2026-01-19
FEAT: Reapply mask to discarded mutations

### LINKED ISSUE #36: Apply mappability filters to the panels

### LINKED ISSUE #53: Create a list / BED file of all the problematic positions, maybe per sample and per cohort...

### LINKED ISSUE #117: Add step to report which mutations have been filtered out
- [ ] Germline variants
- [ ] cohort_n_rich that could be potential hotspots

### LINKED ISSUE #118: Mask sites where mutations have been discarded
If we discard sites based on the `cohort_n_rich` filter, we should also exclude those sites from the randomization.

The best solution to this is to generate a BED file or a list of positions that need to be masked.
The best way to mask them is to put the depth to 0 in the annotate depths step for example

m-huertasp:
## Copilot summary
This pull request introduces a new filtering step for annotated panels to remove variants overlapping flagged genomic regions, updates the panel creation workflow to integrate this filtering, and makes several related configuration and workflow changes. Additionally, some code reviews from [PR153](https://github.com/bbglab/deepCSA/pull/153) are reapplied

**Key changes:**

**New module for filtering captured panels:**
- Added the `filter_captured_panel` process (`modules/local/filter_captured_panel/main.nf`) and its metadata (`modules/local/filter_captured_panel/meta.yml`), which filters annotated panels using `bedtools intersect` to remove variants overlapping flagged regions. Outputs both filtered panels and a list of removed variants. [[1]](diffhunk://#diff-7fc1c94b5719ac539564db4930aafef7d17215144ac2cd2d911f8c72cea01628R1-R51) [[2]](diffhunk://#diff-3def2243522ad0fdfe4322b4aee45fd8fd917434ade6475c9724ef96b098e4e7R1-R50)

**Integration of filtering into panel creation workflow:**
- Refactored the panel creation workflow (`subworkflows/local/createpanels/createpanels/main.nf`, previously `main.nf`) to include the new filtering step (`FILTERCAPTUREDPANEL`). The workflow now takes a `flagged_bed` input from `MUTPREPROCESSING` and generates panels from the filtered annotated panel. The output now includes the list of removed variants. [[1]](diffhunk://#diff-566227d57b6dfe473bd1eb0e238a4776f30e39cc9983fca8386008fc184afd00L1-R18) [[2]](diffhunk://#diff-566227d57b6dfe473bd1eb0e238a4776f30e39cc9983fca8386008fc184afd00R40-R51) [[3]](diffhunk://#diff-566227d57b6dfe473bd1eb0e238a4776f30e39cc9983fca8386008fc184afd00L128-R87)
- Removed the annotation and customization steps from the main panel creation workflow, moving them to a separate annotation workflow (`subworkflows/local/createpanels/annotatepanels/main.nf`). [[1]](diffhunk://#diff-2a4c41ae4839d7e5ab72c0a1c44512375aed76b9b1469fd70ebe02c9fd53a87fR1-R95) [[2]](diffhunk://#diff-566227d57b6dfe473bd1eb0e238a4776f30e39cc9983fca8386008fc184afd00L156-L163)

**Workflow and configuration updates:**
- Updated the main workflow (`workflows/deepcsa.nf`) to use the new, more modular `CREATEPANELS` and `ANNOTATEPANELS` subworkflows.
- Modified the mutation preprocessing workflow to emit the `flagged_bed` output from the `FILTERBATCH` process, enabling downstream filtering.
- Updated the `FILTERBATCH` process in `conf/modules.config` to publish `.bed` files for flagged mutations, and adjusted `MERGEBATCH` and related processes to disable publishing.
- Added output of flagged mutation `.bed` files to the `FILTER_BATCH` process in `modules/local/filtermaf/main.nf`.


modules/local/filter_captured_panel/main.nf
```diff
@@ -0,0 +1,51 @@
...
+
+    container "community.wave.seqera.io/library/bedtools_pybedtools_pandas_pip_pruned:78080da05d53636d"
+
+    input:
+    tuple val(meta), path(complete_annotated_panel)
+    tuple val(meta), path(flagged_bed)
+
+    output:
+    tuple val(meta), path("captured_panel.tab.filtered.tsv")                      , emit: filtered_annotated_panel
+    path("captured_panel.tab.removed_variants.tsv")              , emit: removed_variants
```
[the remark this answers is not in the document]
m-huertasp: Added the output from FILTER_CAPTURED_PANEL into the `createpanels` folder.

subworkflows/local/createpanels/annotatepanels/main.nf
```diff
@@ -0,0 +1,95 @@
...
+
+    panel_annotated_rich           = rich_annotated
+    added_custom_regions           = added_regions
+    domains_panel_bed              = DOMAINANNOTATION.out.domains_bed.first()
+    domains_in_panel               = DOMAINANNOTATION.out.domains_tsv.first()
+
+    postprocessed_panel            = POSTPROCESSVEPPANEL.out.compact_panel_annotation.first()
+    postprocessed_panel_rich       = POSTPROCESSVEPPANEL.out.rich_panel_annotation.first()
+
+    all_consensus_bed_initial    = CREATECONSENSUSPANELSALL.out.consensus_panel_bed.first()
```
FerriolCalvet:
```suggestion
    all_consensus_bed_initial      = CREATECONSENSUSPANELSALL.out.consensus_panel_bed.first()    
```

FerriolCalvet: alignment

workflows/deepcsa.nf
```diff
@@ -15,7 +15,8 @@ include { methodsDescriptionText    } from '../subworkflows/local/utils_nfcore_d
 include { INPUT_CHECK                                       } from '../subworkflows/local/input_check'
 
 include { DEPTH_ANALYSIS            as DEPTHANALYSIS        } from '../subworkflows/local/depthanalysis/main'
-include { CREATE_PANELS             as CREATEPANELS         } from '../subworkflows/local/createpanels/main'
+include { CREATE_PANELS             as CREATEPANELS         } from '../subworkflows/local/createpanels/createpanels/main'
+include { ANNOTATE_PANELS           as ANNOTATEPANELS       } from '../subworkflows/local/createpanels/annotatepanels/main'
```
FerriolCalvet: Very minor change in updating the order
```suggestion
include { ANNOTATE_PANELS           as ANNOTATEPANELS       } from '../subworkflows/local/createpanels/annotatepanels/main'
include { CREATE_PANELS             as CREATEPANELS         } from '../subworkflows/local/createpanels/createpanels/main'
```

subworkflows/local/createpanels/annotatepanels/main.nf
```diff
@@ -0,0 +1,95 @@
...
+    DOMAINANNOTATION(rich_annotated, domains)
+
+    // Create MUTPREPROCESSING-specific panels
+    CREATECAPTUREDPANELS(complete_annotated_panel)
+
+    // All consensus bed
+    restructurePanel(CREATECAPTUREDPANELS.out.captured_panel_all).set{all_panel}
+    if (params.create_sample_panels){
+        CREATESAMPLEPANELSALL(all_panel, depths, params.sample_panel_min_depth)
+    }
```
FerriolCalvet: have you tested the creation of sample panels? this is something that we are not using at all... I wonder if we should just remove it for overall simplicity and reducing the messiness of the code

modules/local/filter_captured_panel/main.nf
```diff
@@ -0,0 +1,51 @@
+process FILTER_CAPTURED_PANEL {
+    tag "$meta.id"
+    label 'process_single'
+    label 'process_medium_high_memory'
+
+    container "community.wave.seqera.io/library/bedtools_pybedtools_pandas_pip_pruned:78080da05d53636d"
+
+    input:
+    tuple val(meta), path(complete_annotated_panel)
+    tuple val(meta), path(flagged_bed)
```
FerriolCalvet:
```suggestion
    tuple val(meta2), path(flagged_bed)
```

modules/local/filter_captured_panel/main.nf
```diff
@@ -0,0 +1,51 @@
+process FILTER_CAPTURED_PANEL {
+    tag "$meta.id"
+    label 'process_single'
+    label 'process_medium_high_memory'
+
+    container "community.wave.seqera.io/library/bedtools_pybedtools_pandas_pip_pruned:78080da05d53636d"
```
FerriolCalvet: it would be good to take the Dockerfile and upload the image to our dockerhub and the dockerfile to our docker github repo

modules/local/filter_captured_panel/main.nf
```diff
@@ -0,0 +1,51 @@
...
+
+    input:
+    tuple val(meta), path(complete_annotated_panel)
+    tuple val(meta), path(flagged_bed)
+
+    output:
+    tuple val(meta), path("captured_panel.tab.filtered.tsv")                      , emit: filtered_annotated_panel
+    path("captured_panel.tab.removed_variants.tsv")              , emit: removed_variants
+
```
FerriolCalvet:
```suggestion
    tuple val(meta), path("captured_panel.tab.filtered.tsv")     , emit: filtered_annotated_panel
    path("captured_panel.tab.removed_variants.tsv")              , emit: removed_variants

```

modules/local/filter_captured_panel/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    head -n 1 ${complete_annotated_panel} | awk '{print \$0 "\\tBED_CHROM\\tBED_START\\tBED_END\\tFILTER"}' > captured_panel.tab.removed_variants.tsv
+
+    # Create temporary bed file from complete annotated panel and filter out flagged regions
+    tail -n +2 ${complete_annotated_panel} | \
+    awk 'BEGIN {FS=OFS="\\t"} {print \$1, \$2-1, \$2, \$0}' > temp_captured_panel.bed
+
+    bedtools intersect -a temp_captured_panel.bed -b ${flagged_bed} -v | cut -f4- >> captured_panel.tab.filtered.tsv
+
+    # Create a separate file with removed variants
+    bedtools intersect -a temp_captured_panel.bed -b ${flagged_bed} -wa -wb | cut -f4- >> captured_panel.tab.removed_variants.tsv
```
FerriolCalvet: I would suggest calling it removed_sites because here we list all the possible mutations in the specific positions where there was the filter

bin/filter_cohort.py
```diff
@@ -1,189 +1,452 @@
...
-    #######
-
+# Logging
+logging.basicConfig(
+    format="%(asctime)s | %(levelname)s | %(name)s - %(message)s", level=logging.DEBUG, datefmt="%m/%d/%Y %I:%M:%S %p"
+)
+LOG = logging.getLogger("filter_cohort")
+
+# Globals
+FILTERS = ["cohort_n_rich", "cohort_n_rich_uni", "other_sample_SNP", "repetitive_variant", "cohort_n_rich_threshold", "gnomAD_SNP", "nanoseq_snp", "nanoseq_noise"]
```
FerriolCalvet: We briefly discussed this off line and I still believe that it would be nice to directly add a new parameter to set this list of filters to apply to the background.

I am removing `cohort_n_rich_threshold` from this list since it is not applied cohort-wise but has some sample specificity. This filter would benefit from the sample specific filtering with the deths (see overall comment)

I am removing `repetitive_variant` since we do not necessarily want to filter those out (this would benefit from having a parameter to indicate this list of filters to be applied to the background)

```suggestion
FILTERS = ["cohort_n_rich", "cohort_n_rich_uni", "other_sample_SNP",  "gnomAD_SNP", "nanoseq_snp", "nanoseq_noise"]
```

bin/filter_cohort.py
```diff
@@ -1,189 +1,452 @@
...
+    maf_df : pandas.DataFrame
+        MAF dataframe with gnomAD SNPs flagged
+    """
+    LOG.info("Flagging gnomAD SNPs...")
+
+    # Flag gnomAD SNPs
     if "gnomAD_SNP" in maf_df.columns:
         maf_df["gnomAD_SNP"] = maf_df["gnomAD_SNP"].replace({"True": True, "False": False, '-' : False}).fillna(False).astype(bool)
-        print("Out of ", maf_df["gnomAD_SNP"].shape[0], "positions", maf_df["gnomAD_SNP"].sum(), "are gnomAD SNPs (>0.1)")
+        LOG.info("Out of %d positions, %d are gnomAD SNPs (>0.1)", maf_df["gnomAD_SNP"].shape[0], maf_df["gnomAD_SNP"].sum())
```
FerriolCalvet: this value can be tuned by the user, maybe we can just leave gnomAD_SNPs to ensure that we are not confusing anyone

bin/filter_cohort.py
```diff
@@ -1,189 +1,452 @@
...
+    required_filters = {"not_covered", "not_in_exons"}
+    all_filters.update(required_filters)
+
+    # Create boolean columns efficiently
+    for filt in sorted(all_filters):
+        maf_df[f"FILTER.{filt}"] = filter_sets.apply(lambda x: filt in x)
+
+    return maf_df
+
+def extract_flagged_regions_bed(maf_df: pd.DataFrame, maf_file_name: str):
```
FerriolCalvet: if we go for the implementation where we apply sample specific filters, we would need to report this information per sample as well

FerriolCalvet (commented):
Looks good Marta!
I added some minor comments in some files but I also have some more general comments that I would like to discuss before merging, mainly because there are different masking scenarios that probably I did not acknowledge properly :pray-hands:

(I am trying to explain what I understood to make sure that I got it properly)
The way it works for the masking is by removing the positions from the panels let's say that there is a flagged position in the middle of an exon, this would result in the exon in the BED regions of the consensus panel being split in two right? and on that position not being in the consensus panel TSV.
OncodriveFML requires a BED file with the coordinates of the different genomic regions, this is probably not impacting at all, but we could look at the consequence of this change outside omega.

Initially I thought that a possibility could be to "just" putting the depths to 0 of that position and then this could have a similar effect to the final definitions of panels. The reality is that with the current implementation this works perfect for positions that need to be masked for the entire cohort, but does not allow the flexibility of removing a position only in one sample.
We can discuss whether we want to merge this implementation and then work on an additional processing that enables the sample-specific tuning of depths that could then impact the panel building process downstream.

Minor more general comments:
* do we want to use the ymls? I guess it does no harm here, but we should discuss it, and probably apply it as part of the clean branch

m-huertasp:
Thank you for the review Ferriol! I will check the comments tomorrow, but I wanted to answer the general comment so everything is here in case we need it.

You're right, removing a position from the panel makes the regions in the bed to be split. 
Regarding the option of just forcing the depths to be 0 for a position, I thought that this option was discarded due to the problems when computing omega and other values using the depth as denominator, I assume this is still a problem, right?. I was following Fede's suggestion in the issue, although it is true that this cannot be applied per sample.
Maybe we could try the option of forcing the depths to be 0 in each specific sample depth when the position is masked. Then add a step in which all the depths = 0 are removed. Could this be a good idea? Maybe we could combine this with the approach we have here so we avoid removing the same positions in all the samples each time?

About the ymls, I think it could be useful for new people trying to understand how the pipeline works. But I am okay with moving them to the clean branch.

FerriolCalvet:
Some time ago I made some fixes to prevent the 0s problem Fede experienced above (https://github.com/bbglab/deepCSA/commit/4551e178c6cc1b0f9699a6c16d358acaa735137f), so hopefully this should not be a problem now. I don't remember if there were other methods for which this was also a problem but I think it is worth testing it again.
I would try to avoid having to filter positions from panels again after this first masking steps, but we can see how things work.

And yes, agree with the ymls, I would leave the ones you created here and then try to add updated ones for all steps in the clean branch.

let's discuss soon!

subworkflows/local/createpanels/annotatepanels/main.nf
FerriolCalvet: we agree on removing it

modules/local/filter_captured_panel/main.nf
FerriolCalvet: revise where else it is used and maybe replace everywhere

FerriolCalvet: it is in createcapturedpanels

bin/filter_cohort.py
FerriolCalvet: we agree on using the filter_criteria and filter_criteria_somatic to define which of these flags should be used for masking the panels

m-huertasp:
We have decided to try applying the option of forcing the depth to 0 for those positions to be removed in specific samples. If it is too complex, we could merge this branch and do it later.

bin/filter_cohort.py
m-huertasp: So, just to clarify and make this right. This are the default values of the filters:
```nextflow
filter_criteria = ["notcontains NM20", "notcontains n_rich", "notcontains cohort_n_rich_threshold", "notcontains cohort_n_rich", "notcontains no_pileup_support", "notcontains low_mappability", "notcontains not_covered", "notcontains nanoseq_noise"]
filter_criteria_somatic = ["notcontains gnomAD_SNP", "notcontains nanoseq_snp"]
```
I was thinking that we could use this same parameters and let the user define which positions to filter depths/panels as well (which will be the ones in the mutations), but using the sample-specific parameters in the depth=0 filtering and the others in this approach. What do you think, is this what you were thinking?

FerriolCalvet: If I understand it properly yes, basically all the filters that are applied should be removed from the background, either at cohort-level or sample-level.

m-huertasp: I have added the `filter_criteria` and `filter_criteria_somatic` as inputs to `filter_cohort.py.` In there, there is a list of "cohort specific filters", only the intersection of those filters and the ones in the parameters are masked at this step.

m-huertasp:
## Update
**About the depth = 0 mask**
What I have done is to create a new module SAMPLEFLAGGEDBED in which a bed file with the flagged positions is created per sample. I know it is not ideal to have more modules, the other option I was thinking was to do the generation of sample beds in the MERGEBATCH step, but this would not be parallelized. For this step, I also use the `filter_criteria`, but excluding those filters that are masked at cohort level.

**The problem**
The only step in which I think we can mask the depths is in the ANNOTATEDEPTHS. This step gets a file with the depths of all the samples and outputs the depths of each sample and per groups. So you cannot parallelize here the sample masking. I have different options.
- Option 1: We could add a loop in the `merge_annotation_depths.py` so it gets a path for the bed files and it filters each sample, but I think it is not super efficient.
- Option 2: We could mask before ANNOTATEDEPTHS, this will be good because different concerns will be separated but not super efficient either and adds one more module.
- Option 3: We get all the beds and create a "mask matrix" with the positions as rows and the samples as columns, TRUE if the position in that sample is masked, FALSE if it is not. We could add this mask matrix to ANNOTATEDEPTHS and do the filtering just once. My doubts about this are related with the memory used and such.

I like option 3. Do you have any thoughts about this?

FerriolCalvet:
I was thinking on something like option 3 as well! I understand the memory concerns but the good thing is that since the masking of each position is independent from the others the processing of the matrix could be split in chunks if needed. But for now I would not worry about it and we can test it in the duplexomes once you have a working version of it for a smaller panel such as the bladder one.

m-huertasp:
## Update
I have successfully applied the depth = 0 mask. The depths are mask correctly and the pipeline seems to be running without problems (omega, OncodriveFML, etc). I will write here some details of how it is done so we have it all written.

**Panel filtering approach**
- I have removed all the changes related to the "panel filtering" strategy. The CREATEPANELS works as it was working before. It does not receive any input from MUT_PREPROCESSING.

**Depth = 0 approach**
Most of the changes are added at the MUT_PREPROCESSING step:
- I have added a new module **SAMPLEFLAGGEDBED**. It gets the maf files per sample already with the FILTERS. Generates one bed file (1-based coordinates as the depth files) per sample. This module uses the `filter_criteria` and `filter_criteria_somatic` to know which positions to add to the bed file.
   - **PROBLEM:** Some FILTERS are applied later at the cohort level in **FILTERBATCH**. In this step the maf files are merged into one and some more filters (cohort_n_rich, repetitive_variant, gnomAD_SNP, etc) are added.
- To solve this I am getting a bed file from the **FILTERBATCH** step as well. This bed file takes into account only the intersection between those FILTERS added in this step and the ones added by the user as `filter_criteria` and `filter_criteria_somatic`.  
- The bed files at cohort and sample levels go into a new output folder named `flagged_positions`.
- All the sample-specific and "cohort" bed files go into another new module called **CREATEMASKMATRIX**. This returns a matrix with positions as rows and samples as columns where there is a 1 if the position should be kept and a 0 if the position should be masked.
- The **ANNOTATEDEPTHS** step now uses the `mask_matrix` to filter sample positions, assigning a depth of 0 where the mask value is zero.

bin/merge_annotation_depths.py
```diff
@@ -1,78 +1,199 @@
...
+            annotated_depths.select(COLS + [str(sample)]).write_csv(temp_file, separator="\t", include_header=True)
+            subprocess.run(["gzip", "-f", temp_file], check=True)
+
+        # All the samples
+        annotated_depths = annotated_depths.with_columns(
+            pl.sum_horizontal(pl.exclude(COLS)).alias("all_samples"))
+        
+        temp_file = "all_samples.depths.annotated.tsv"
+        annotated_depths.select(COLS + ["all_samples"]).write_csv(temp_file, separator="\t", include_header=True)
+        subprocess.run(["gzip", "-f", temp_file], check=True)
```
FerriolCalvet, line 179: I am not sure if there is any scenario where we would not have the JSON file, so maybe this part is not needed, but let's revise just in case

bin/utils_filter.py
```diff
@@ -0,0 +1,128 @@
...
+        List of filter names to apply
+    """
+    # Parse comma-separated strings into lists
+    filter_list = [f.strip() for f in filters.split(',') if f.strip()]
+    somatic_filter_list = [f.strip() for f in somatic_filters.split(',') if f.strip()]
+    
+    # Combine both lists
+    all_filters = filter_list + somatic_filter_list
+
+    result = [f.replace("notcontains ", "") for f in all_filters if f.startswith("notcontains ")]
```
FerriolCalvet, line 93: there might be other definitions of filters, but for now it is true that this covers most of the cases that we actually use

see filter_maf function in utils.py for the complete list of possibilities

As a side note, maybe we can move that function here since it is related to filters

conf/tools/panels.config
```diff
@@ -23,6 +23,20 @@ process {
...
+        publishDir = [
+            enabled: false
+        ]
+    }
+
+    withName: 'BBGTOOLS:DEEPCSA:CREATEPANELS:CREATECAPTUREDPANELS.*' {
+        publishDir = [
+            enabled: false
+        ]
+    }
```
FerriolCalvet: I am not sure that this is needed any more

conf/modules.config
```diff
@@ -189,7 +189,7 @@ process {
         ext.underscore          = false
     }
 
-    withName: 'FILTERBATCH|MERGEBATCH|VCF2MAF|COMPUTETRINUC|COMPUTEDEPTHS' {
+    withName: 'FILTERBATCH|MERGEBATCH|VCF2MAF|COMPUTETRINUC|COMPUTEDEPTHS|CREATEMASKMATRIX' {
```
FerriolCalvet: I think maybe it is worth outputting the matrix, just to not lose this summarized view of the filters

conf/modules.config
```diff
@@ -316,18 +324,11 @@ process {
...
 
-
-    withName: FILTERBATCH {
-        ext.repetitive_variant = params.repetitive_variant_thres
-        ext.germline_threshold = params.germline_threshold
-    }
-
-
-
     withName: "TABLE2GROUP" {
```
FerriolCalvet, line 328: I understand that as it was implemented this information was not used at all, however in general we tend to define all these variables here and then load them in the script, so I think that for consistency I would prefer to have all the parameters listed here instead of having them in the script

conf/modules.config
```diff
@@ -316,18 +324,11 @@ process {
         }
     }
 
+
     withName: 'DOWNSAMPLE.*' {
         ext.downsample_prop = params.downsample_proportion
     }
```
FerriolCalvet, line 321:
```suggestion
    }

    withName: FILTERBATCH {
        ext.repetitive_variant = params.repetitive_variant_thres
        ext.germline_threshold = params.germline_threshold
        ext.prop_samples_nrich = params.prop_samples_nrich
    }
```
and then modify the nf script to handle this appropriately and if there is any missing value of this parameters, then do not add the flag to the command

modules/local/createmaskmatrix/main.nf
```diff
@@ -0,0 +1,37 @@
+process CREATE_MASK_MATRIX {
+    tag "mask_matrix"
+
+    label 'process_low'
+    label 'time_low'
+
+    container "docker.io/bbglab/deepcsa-core:0.0.2-alpha"
```
FerriolCalvet: need to update the container to the last one

modules/local/createmaskmatrix/main.nf
```diff
@@ -0,0 +1,37 @@
...
+
+    input:
+    path(bed_files)  // List of sample-specific flagged position BED files
+
+    output:
+    path("flagged_positions.mask.tsv.gz")          , emit: mask_matrix
+    path "versions.yml"                            , topic: versions
+
+    script:
+    def bed_args = bed_files.collect { "\"${it}\"" }.join(' --bed-files ')
```
FerriolCalvet, line 17:
```suggestion
    def bed_args = bed_files.collect { it -> "\"${it}\"" }.join(' --bed-files ')
```

modules/local/writemaf/main.nf
```diff
@@ -27,7 +33,9 @@ process WRITE_MAFS {
 
     stub:
     """
-    touch all_samples.cohort.tsv.gz
+    touch all_samples.filtered.tsv.gz
+    touch all_samples.flagged-pos.bed
+    touch *.flagged-pos.bed
```
FerriolCalvet, line 40: I don't this this would work no?, but anyway we can forget about it

modules/local/writemaf/main.nf
```diff
@@ -10,14 +10,20 @@ process WRITE_MAFS {
     path (json_groups)
 
     output:
-    path("*.filtered.tsv.gz") , emit: mafs
-    path "versions.yml"       , topic: versions
+    path("*.filtered.tsv.gz")              , emit: mafs
+    path("all_samples.flagged-pos.bed")    , emit: all_samples_flagged_positions
```
FerriolCalvet: I don't understand why this one is outputted separately? I think it is not used anywhere

bin/utils_filter.py
```diff
@@ -0,0 +1,128 @@
...
+    Parse filter criteria from comma-separated strings.
+    
+    Parameters
+    ----------
+    filters : str
+        Comma-separated list of filter criteria
+    somatic_filters : str
+        Comma-separated list of somatic filter criteria
+    only_cohort_filters : bool
+        If True, only include cohort-level filters from the returned list
```
FerriolCalvet: this parameter is not there any more, but it could be interesting to recover so that this function returns two lists, one with all the filters and another one only with cohort level filters that would be the intersection with the list of filters that are defined at cohort level

bin/write_mafs.py
```diff
@@ -1,29 +1,121 @@
...
     for group_name, samples in groups_info.items():
         samples = [str(x) for x in samples]
-        maf_df[maf_df["SAMPLE_ID"].isin(samples)].sort_values(by=["CHROM", "POS"]).to_csv(
-            f"{group_name}.filtered.tsv.gz",
-            sep="\t",
-            header=True,
-            index=False
-        )
+        # Extract group-specific MAFs
+        create_group_mafs(maf_df, group_name, samples)
```
FerriolCalvet, line 126: I understand that this is the exact same functionality as before right?

bin/utils_filter.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+    LOG.info("Mutations flagged: %s", _bed_melt.shape[0])
+
+    # Aggregate filters per position
+    bed_annotated = (
+                _bed_melt
+                .drop_duplicates()
+                .groupby(["CHROM","POS"])["FILTERS"]
+                .agg(','.join)
```
FerriolCalvet, line 175: it will probably already be sorted, but otherwise you could also sort before joining it is it not too difficult, just to ensure consistency in the outputs

bin/write_mafs.py
```diff
@@ -1,29 +1,121 @@
...
+
+    # Extract flagged regions for each sample in the group if not already done
+    create_sample_flagged_bed(maf_df, all_samples_bed, filter_criteria)
+
+    # Extract flagged regions to BED file (cohort-wide, applies to all samples)
+    LOG.info("Extracting cohort-wide flagged positions to BED file.")
+    extract_flagged_regions_bed(maf_df, "all_samples", filter_criteria)
```
FerriolCalvet: here is where we could use only the set of filters that are applied in our filtering criteria and that are cohort-level, since when we report a position is n_rich this means that it is n_rich in a sample but not necessarily removed from all the samples in the cohort

bin/filter_cohort.py
```diff
@@ -1,189 +1,361 @@
...
+        LOG.warning("Not enough samples to identify potential repetitive variants!")
+
+        return maf_df
+
+    # Work with already filtered df + somatic only to explore potential artifacts
+    # take only variant and sample info from the df
+    maf_df_f_somatic = maf_df.loc[maf_df["VAF"] <= somatic_vaf_boundary][["MUT_ID","SAMPLE_ID", "PMEAN", "PSTD"]].reset_index(drop = True)
+
+    # Group by 'MUT_ID' and count occurrences
+    maf_df_f_somatic_pivot = maf_df_f_somatic.groupby("MUT_ID").size().reset_index(name="count")
```
FerriolCalvet, line 93: it was so cutre, thanks!

bin/filter_cohort.py
```diff
@@ -1,189 +1,361 @@
...
-    ###  Filter cohort_n_rich
-    #######
+    if n_samples == 0:
+        LOG.warning("Not enough samples to identify potential repetitive variants!")
+
+        return maf_df
+
+    # Work with already filtered df + somatic only to explore potential artifacts
+    # take only variant and sample info from the df
+    maf_df_f_somatic = maf_df.loc[maf_df["VAF"] <= somatic_vaf_boundary][["MUT_ID","SAMPLE_ID", "PMEAN", "PSTD"]].reset_index(drop = True)
```
FerriolCalvet, line 90: I think in other places the definition of somatic is made filtering also for the vd_VAF and VAF_AM columns, maybe we should uniformize it, but probably it is out of the scope of this PR and we can leave it for later

bin/filter_cohort.py
```diff
@@ -1,189 +1,361 @@
...
+    
+    number_of_samples = max(2, (max_samples * n_rich_cohort_proportion) // 1)
+    LOG.info(f"Flagging mutations that are n_rich in at least: {number_of_samples} samples as cohort_n_rich")
+
+    # Work with already filtered df + somatic only to explore potential artifacts
+    # take only variant and sample info from the df.
+    maf_df_f_somatic = maf_df[["MUT_ID", "SAMPLE_ID", "VAF_Ns", "FILTER"]].reset_index(drop = True)
+    
+    # Note: Pending discussion on whether to filter by somatic VAF boundary here or not
+    # maf_df_f_somatic = maf_df.loc[maf_df["VAF"] <= somatic_vaf_boundary][["MUT_ID", "SAMPLE_ID", "VAF_Ns", "FILTER"]].reset_index(drop = True)
```
FerriolCalvet: uff I see that it was called somatic but it was actually not somatic, I think it is OK to do it will all variants no only somatic ones but then maybe we should rename the variables in this function.

thanks for noticing!

bin/filter_cohort.py
```diff
@@ -1,189 +1,361 @@
...
-            maf_df[f"FILTER.{filtt}"] = False
-
-
+    
+    maf_df = maf_df.drop("gnomAD_SNP", axis = 1)
+
+    return maf_df
+
+
+def flag_maf(maf_df: pd.DataFrame, sample_name: str,
```
FerriolCalvet, line 343: thanksss much cleaner now, even if there are still some issues with the old code, we can fix them later

bin/utils_filter.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # The idea is to filter depth files at these positions, so make END = START (1-based)
+    bed_annotated["END"] = bed_annotated["START"]
+
+    LOG.info("Unique regions flagged: %s", bed_annotated.shape[0])
+
+    # Write BED file without header or index
+    (bed_annotated[["CHROM", "START", "END", "FILTERS"]]
+        .sort_values(by=["CHROM", "START"])
+        .to_csv(f"{name}.flagged-pos.bed", sep="\t", header=False, index=False)
```
FerriolCalvet: call it a different name so that it does not go in the same channel as the other BED files and then you will avoid some problems downstream
```suggestion
        .to_csv(f"{name}.cohort-flagged-pos.bed", sep="\t", header=False, index=False)
```

modules/local/writemaf/main.nf
FerriolCalvet:
```suggestion
    path("all_samples.cohort-flagged-pos.bed")    , emit: all_samples_flagged_positions
```

bin/create_mask_matrix.py
```diff
@@ -0,0 +1,196 @@
...
+    Parameters
+    ----------
+    bed_files : list
+        List of paths to sample-specific BED files
+    """
+    LOG.info(f"Processing {len(bed_files)} BED files...")
+    
+    # Filter out all_samples aggregated file if present
+    sample_bed_files = [f for f in bed_files if 'all_samples' not in Path(f).stem]
+
```
FerriolCalvet: applying some suggested fixes to the write mafs script this will not be needed anymore

FerriolCalvet: by this I mean the exclusion of the all_samples files

FerriolCalvet (commented):
looks great!!!
There are very minor things but I believe that once we merge this branch then we can go to tackle improvements in the namings of processes, variables and many things

The only change I think is worth applying and that requires a bit more of work is that the all samples flagged positions only includes the ones filtered in the entire cohort, or we could have one that includes all and another with only the cohort-wide ones. This could help inform us about positions that are to be discarded always

bin/merge_annotation_depths.py
m-huertasp: This refactoring was in fact done by Fede (I just reapplied it with `polars`), it is doing the same as the original code, if there is a JSON with the group, generate the depth file per group and then per sample. If there is no JSON file generate it only per sample. 

Other change there is the creation of temporal files that are gzipped in a subprocess, this is an annoying thing that I didn't know about `polars`. It seems that it is quite problematic to output a gzip csv with `polars`. This is a way I found to do it. Another way is tranforming to `pandas` but then you need to install a different library...

bin/write_mafs.py
m-huertasp: Yess, I just moved it to a function outside to have this "main" a bit cleaner, but we could move it here if you prefer

bin/filter_cohort.py
m-huertasp: I've added a new issue #418

modules/local/createmaskmatrix/main.nf
m-huertasp: This didn't work :(

FerriolCalvet: ohh okay then forget haha

bin/merge_annotation_depths.py
FerriolCalvet: okay... well it looks good to me this way as well if it works, but if you want to try something else feel free to do it now or we leave it for later

modules/local/filtermaf/main.nf
FerriolCalvet, line 1: perfect! much cleaner now, thanks

FerriolCalvet (approved):
let's mergeee!!

bin/merge_annotation_depths.py
```diff
@@ -1,81 +1,200 @@
...
-    input_csv = pd.read_table(input_file, sep = ',', header = 0)
-    if 'bam' in input_csv.columns:  
-        bam_basenames = input_csv["bam"].astype(str).str.split("/").str[-1]  
-        bam2sample_dict = dict(zip(bam_basenames, input_csv["sample"].astype(str)))  
+    # Place COLS=[CHROM, POS, CONTEXT] columns at the beginning then the rest of the columns
+    return annot_depth.select(COLS + sample_columns).rename(rename_map), list(rename_map.values())
 
-    else:  
-        samples = input_csv["sample"].astype(str)  
-        bam2sample_dict = dict(zip(samples, samples))
```
FerriolCalvet, line 34: @m-huertasp 
we should get this renaming here back in, let's discuss how now in the lab

PR #420, opened 2026-02-17
Fix no filters usecase

FerriolCalvet:
This PR implements some changes to fix a bug created by the updates regarding the reporting of positions to be masked and so on.

## AI summary
This pull request improves the handling of optional filter parameters and output files in the `WRITE_MAFS` process. The changes ensure that filters are only applied when specified and that certain output files are treated as optional, making the workflow more robust to missing or empty inputs.

Parameter handling improvements:

* Updated the assignment of `ext.filters` and `ext.somatic_filters` in `conf/modules.config` to set them to an empty string if the corresponding parameter is not provided, preventing errors from attempting to join a null value.
* Modified the construction of the `filters` and `somatic_filters` command-line arguments in `modules/local/writemaf/main.nf` to only include these options if the corresponding values are set, avoiding unnecessary or empty arguments in the script call.

Output file handling improvements:

* Marked the output files `all_samples.all-flagged-pos.bed` and `all_samples.cohort-wide-flagged-pos.bed` as optional in the `WRITE_MAFS` process, so the workflow will not fail if these files are not generated.

### DIFF
--- a/modules/local/writemaf/main.nf
+++ b/modules/local/writemaf/main.nf
@@ -10,11 +10,11 @@ process WRITE_MAFS {
     path (json_groups)
 
     output:
-    path("*.filtered.tsv.gz")                                       , emit: mafs
-    path("*.flagged-pos.bed")                                       , emit: sample_flagged_beds
-    path("all_samples.all-flagged-pos.bed")         , optional: true
-    path("all_samples.cohort-wide-flagged-pos.bed") , optional: true
-    path "versions.yml"                         , topic: versions
+    path("*.filtered.tsv.gz")                       , emit: mafs
+    path("*.flagged-pos.bed")                       , emit: sample_flagged_beds
+    path("all_samples.all-flagged-pos.bed")         
+    path("all_samples.cohort-wide-flagged-pos.bed") 
+    path "versions.yml"                             , topic: versions
 
     script:
     def filters = task.ext.filters ? "--filters \"${task.ext.filters}\"" : ""
--- a/modules/local/writemaf/main.nf
+++ b/modules/local/writemaf/main.nf
@@ -10,10 +10,10 @@ process WRITE_MAFS {
     path (json_groups)
 
     output:
-    path("*.filtered.tsv.gz")                   , emit: mafs
-    path("*.flagged-pos.bed")                   , emit: sample_flagged_beds
-    path("all_samples.all-flagged-pos.bed")
-    path("all_samples.cohort-wide-flagged-pos.bed")
+    path("*.filtered.tsv.gz")                                       , emit: mafs
+    path("*.flagged-pos.bed")                                       , emit: sample_flagged_beds
+    path("all_samples.all-flagged-pos.bed")         , optional: true
+    path("all_samples.cohort-wide-flagged-pos.bed") , optional: true
     path "versions.yml"                         , topic: versions
 
     script:
--- a/modules/local/writemaf/main.nf
+++ b/modules/local/writemaf/main.nf
@@ -17,14 +17,14 @@ process WRITE_MAFS {
     path "versions.yml"                         , topic: versions
 
     script:
-    def filters = task.ext.filters ?: ""
-    def somatic_filters = task.ext.somatic_filters ?: ""
+    def filters = task.ext.filters ? "--filters \"${task.ext.filters}\"" : ""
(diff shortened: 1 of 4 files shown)


modules/local/writemaf/main.nf
```diff
@@ -10,21 +10,21 @@ process WRITE_MAFS {
...
 
     output:
-    path("*.filtered.tsv.gz")                   , emit: mafs
-    path("*.flagged-pos.bed")                   , emit: sample_flagged_beds
-    path("all_samples.all-flagged-pos.bed")
-    path("all_samples.cohort-wide-flagged-pos.bed")
+    path("*.filtered.tsv.gz")                                       , emit: mafs
+    path("*.flagged-pos.bed")                                       , emit: sample_flagged_beds
+    path("all_samples.all-flagged-pos.bed")         , optional: true
+    path("all_samples.cohort-wide-flagged-pos.bed") , optional: true
```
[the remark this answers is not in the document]
m-huertasp: I think this comment could be important. In the `utils_filter.py` line 153 an empty bed file is generated when there are no regions to filter. In there, there is a wrong thing:

Currently it is:
``` python
Path(f"{name}.flagged-pos.bed").touch()
```

It should be:
``` python
Path(f"{name}.{specification}flagged-pos.bed").touch()
```

This specification is only used whan generating the all_samples.all and all_samples.cohort-wide bed files. By adding it you make sure that, when there are no filters, the empty file generated has the same name as the non-empty file and you make sure there is always an output.

modules/local/writemaf/main.nf
```diff
@@ -10,21 +10,21 @@ process WRITE_MAFS {
     path (json_groups)
 
     output:
-    path("*.filtered.tsv.gz")                   , emit: mafs
-    path("*.flagged-pos.bed")                   , emit: sample_flagged_beds
-    path("all_samples.all-flagged-pos.bed")
-    path("all_samples.cohort-wide-flagged-pos.bed")
+    path("*.filtered.tsv.gz")                                       , emit: mafs
+    path("*.flagged-pos.bed")                                       , emit: sample_flagged_beds
+    path("all_samples.all-flagged-pos.bed")         , optional: true
```
m-huertasp: This bed files are only used by `CREATEMASKMATRIX`, right? If so, it should not be a problem for them to be empty. If they are empty the mask matrix will be empty. If the mask matrix is empty, no filtering is done at `ANNOTATEDEPTHS`.

So I think there should not be needed for this bed files to be optional. Was there some error with this?

modules/local/writemaf/main.nf
```diff
@@ -10,21 +10,21 @@ process WRITE_MAFS {
...
+    path("all_samples.cohort-wide-flagged-pos.bed") , optional: true
     path "versions.yml"                         , topic: versions
 
     script:
-    def filters = task.ext.filters ?: ""
-    def somatic_filters = task.ext.somatic_filters ?: ""
+    def filters = task.ext.filters ? "--filters \"${task.ext.filters}\"" : ""
```
m-huertasp, line 20: Nice!

m-huertasp (commented):
Thank you for taking a look at this and sorry for creating this problem! I think you've done nice with the filters, I didn't think about the possibility of having them empty.

Regarding the output of the bed files from WRITEMAFs being optional, I think there could be an easier solution. I think this is a problem caused by me in fact. I've already added what could be the problem in the comments. 

If you have any question or need any help just tell me!

Nice job!

modules/local/writemaf/main.nf
FerriolCalvet: I think that the file was not being generated so then it raised some errors and that's why I added the optional thing, but if you think that the file should still be generated in the absence of any flagged positions we may be able to remove the optional here.
I can test it tomorrow and see

m-huertasp: Yess, I think it should generate an empty file if there are no flagged regions. The problem now is the one in the first comment. When the empty file is created, it has a different name because I forgot to add a part in there, this is why it is giving the no output errors.

> In the utils_filter.py line 153 an empty bed file is generated when there are no regions to filter. In there, there is a wrong thing:
Currently it is: 
Path(f"{name}.flagged-pos.bed").touch()
It should be:
Path(f"{name}.{specification}flagged-pos.bed").touch()

FerriolCalvet: ah okay understood, I will apply this change later today or tomorrow then and test again

thankssss!!!

PR #422, opened 2026-02-24
REFACTOR: Improve dna2protein bottleneck

### LINKED ISSUE #229: use panel annotation to define already the protein position of each CDS position
this might replace the protein2dna mapping step and make it independent from the connections to Ensembl

m-huertasp:
## Summary
This PR optimizes the `DNA2PROTEINMAPPING` step by reducing its bottleneck. Previously, the pipeline executed multiple serial requests to Ensembl REST APIs (roughly 2× the number of genes in a panel), which frequently led to rate-limiting blocks and significant latency for large panels.

Based on suggestions from @FerriolCalvet , the individual calls were replaced with a bulk-retrieval strategy. We now download the full GFF annotation via FTP and process it locally.

### Main Changes
- **New logic**: Replaced multiple REST API calls with a single FTP parsing of the Ensembl GFF. The document is not downloaded but filtered on the fly and briefly kept in memory until transformation into a dataframe.
- **Refactor**: Updated downstream functions to handle DataFrames instead of the JSON objects previously returned by the API.
- **Flexibility**: Linked the Ensembl release version to the vep_cache_version Nextflow parameter, allowing users to specify the genomic release dynamically.
- **Documentation**: Added docstrings and internal comments to explain the new and old parsing logic.

closes #229

### DIFF
--- a/bin/panels_computedna2protein.py
+++ b/bin/panels_computedna2protein.py
@@ -85,7 +85,7 @@ def get_transcript_gene_from_maf(path_maf, consensus_file):
     return gene_transcript_pairs
 
 # Generator to filter lines before they reach a dataframe
-def get_gff_to_generator(release: int = 111):
+def get_gff_to_generator(release: int = 111, species: str = "homo_sapiens", genome: str = "GRCh38"):
     """
     Get GFF file from ensembl FTP and filter it on the fly to keep only exon and CDS lines.
 
@@ -99,7 +99,7 @@ def get_gff_to_generator(release: int = 111):
     generator
         A generator that yields lines from the GFF file that correspond to exon and CDS features.
     """
-    url = f"https://ftp.ensembl.org/pub/release-{release}/gff3/homo_sapiens/Homo_sapiens.GRCh38.{release}.gff3.gz"
+    url = f"https://ftp.ensembl.org/pub/release-{release}/gff3/{species}/{species.capitalize()}.{genome}.{release}.gff3.gz"
 
     # Open request
     try:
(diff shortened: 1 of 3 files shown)


bin/panels_computedna2protein.py
```diff
@@ -27,179 +81,268 @@ def get_transcript_gene_from_maf(path_maf, consensus_file):
...
+    ------------
+    release : int
+        The release number of the Ensembl GFF file to retrieve (default is 111).
 
+    Returns
+    ------------
+    generator
+        A generator that yields lines from the GFF file that correspond to exon and CDS features.
+    """
+    url = f"https://ftp.ensembl.org/pub/release-{release}/gff3/homo_sapiens/Homo_sapiens.GRCh38.{release}.gff3.gz"
```
FerriolCalvet: we might use this script for mice as well, we should fix this to accept mouse also version 111 (and assembly should be already mm39 by default)

FerriolCalvet: GRCm39

conf/modules.config
```diff
@@ -72,6 +72,10 @@ process {
         ]
     }
 
+    withName: DNA2PROTEINMAPPING {
+        ext.ensembl_release = params.vep_cache_version ?: 111
```
FerriolCalvet, line 76: you will have to add the assembly here

FerriolCalvet (changes requested):
all good! great Marta!

modules/local/dna2protein/main.nf
FerriolCalvet, line 1: when testing this file needs to have more memory, particularly for bigger panels, add the process_medium_high_memory label and then merge

FerriolCalvet (commented):
very minor changes from testing

modules/local/dna2protein/main.nf
m-huertasp: This will not be added in this branch but in `ferriol-updates` branch.

PR #424, opened 2026-02-25
plot depths per group

efigb:
This branch adds a new module to process analysis of consensus exons depth for a defined group of samples and for a subset of genes (default genes in the panel) plotting all these possibilities in an output pdf that is stored in the same output directory as plotting the depths (`/depths/summary`)

### DIFF
--- a/bin/depth_group_comparison.py
+++ b/bin/depth_group_comparison.py
@@ -37,7 +37,7 @@ def plot_depth_per_group(df, group_col, data_type, pdf):
     col_name = group_col[0] if isinstance(group_col, list) else group_col
 
     # Get the number of unique categories to plot
-    num_categories = df[col_name].nunique() + 1
+    num_categories = df[col_name].nunique() + 2
     plt.figure(figsize=(num_categories, 4))
 
     ax = sns.boxplot(data=df, x=col_name, y="MEAN_GENE_DEPTH", hue=col_name, showfliers=False, showmeans=False,legend=False)
--- a/bin/depth_group_comparison.py
+++ b/bin/depth_group_comparison.py
@@ -1,0 +1,144 @@
+#!/usr/bin/env python
+
+"""
+Plot depth of all genes and specific genes per sample groups.
+
+This script plots the average depth at all consensus exons across all genes and for specific genes of interest, either in the panel or in a custom subset of genes, stratified by sample groups defined in the metadata.
+The output is stored in a pdf file.
+"""
+
+import click
+import pandas as pd
+from utils_plot import plots_general_config
+import matplotlib.pyplot as plt
+import matplotlib as mpl
+from matplotlib.backends.backend_pdf import PdfPages
+import seaborn as sns
+
+mpl.rcParams.update({
+    'axes.titlesize': plots_general_config["title_fontsize"],
+    'axes.labelsize': plots_general_config["xylabel_fontsize"],
+    'xtick.labelsize': plots_general_config["xyticks_fontsize"],
+    'ytick.labelsize': plots_general_config["xyticks_fontsize"],
+    'figure.titlesize': plots_general_config["title_fontsize"],
+})
+
+separator2character = {
+    'tab' : '\t',
+    'comma' : ','
+}
+
+
+def plot_depth_per_group(df, group_col, data_type, pdf):
+    '''
+    Function to plot depth within a group of samples in all genes or a specific subset of genes
+    '''
+ 
+    col_name = group_col[0] if isinstance(group_col, list) else group_col
+
+    # Get the number of unique categories to plot
+    num_categories = df[col_name].nunique() + 1
+    plt.figure(figsize=(num_categories, 4))
+
+    ax = sns.boxplot(data=df, x=col_name, y="MEAN_GENE_DEPTH", hue=col_name, showfliers=False, showmeans=False,legend=False)
+    ax = sns.stripplot(data=df, x=col_name, y="MEAN_GENE_DEPTH", color='grey', alpha=0.5, size=4, legend=False)    
+
+    if data_type == 'all_genes':
+        plt.title(f"Average Depth for {data_type} in {col_name} group", fontsize=plots_general_config["title_fontsize"])
+    elif data_type == 'gene':
+        gene = df['GENE'].iloc[0]
+        plt.title(f"Average Depth for {gene} in {col_name} group", fontsize=plots_general_config["title_fontsize"])
+    else:
+        print(f"Unknown data type: {data_type}. Title will not be set.")
+    
+    plt.xlabel('', fontsize=plots_general_config["xylabel_fontsize"])
+    plt.ylabel(f"Average Cons Exons Depth", fontsize=plots_general_config["xylabel_fontsize"])
+    plt.yticks( fontsize=plots_general_config["yticks_fontsize"])
+    plt.xticks(fontsize=plots_general_config["xticks_fontsize"])
+    plt.tick_params(axis='x', rotation=90)
+    plt.tight_layout()
+    pdf.savefig()
+    plt.close()
+    plt.show()
+
+    return 
+
+
+@click.command()
+@click.option('--table-filename', required=True, type=click.Path(exists=True), help='Input features table file')
+@click.option('--depth-table', required=True, type=click.Path(exists=True), help='Input depth table file')
+@click.option('--separator', required=True, type=click.Choice(['tab', 'comma']), help='Separator: tab or comma')
+@click.option('--unique-identifier', default=None, type=str, help='Unique identifier column name')
+@click.option('--groups', default=None, type=str, help='List of columns with grouping information')
+@click.option('--custom-genes', required=False, type=str, help='Comma separated list of custom genes')
+@click.option('--output_prefix', type=str, required=True, help='Prefix for output files')
+
+
+def main(table_filename, depth_table, unique_identifier, separator, groups, custom_genes, output_prefix):
+
+    sep_char = separator2character[separator]
+    
+    # Read tables
+    features_table = pd.read_table(table_filename, header=0, sep=sep_char)
+    depth_table = pd.read_table(depth_table, header=0, sep=sep_char)
+
+    # Process panel genes 
+    panel_genes = sorted(set(depth_table['GENE'].unique()))
+    if custom_genes:
+        print(f'Custom genes provided, plotting custom genes only: {custom_genes}')
+        custom_gene_list = [g.strip() for g in custom_genes.split(",")]
+        panel_genes = sorted(set(custom_gene_list) & set(panel_genes))
+    
+    else:
+        print(f'No custom genes provided, plotting all genes in the panel: {panel_genes}')
+
+    output_pdf_path = f"{output_prefix}.plot_depth_per_group.pdf"
+
+    # groups may contain lists of lists, but all formatted into a string
+    groups_of_interest_init = [group.strip().strip(",").split(",") for group in groups.replace("[", ";;;").replace("]", "").split(";;;")] if groups else []
+
+    groups_of_interest = []
+    for comparison in groups_of_interest_init:
+        comparison_group_clean = [item.strip() for item in comparison]
+        comparison_group = [item for item in comparison_group_clean if item != '']
+        if len(comparison_group) > 0:
+            groups_of_interest.append(comparison_group)
+
+    uniq_name = unique_identifier if unique_identifier else "sample"
+
+    print(f"Processing data for the groups of interest: {groups_of_interest}")
+
+    with PdfPages(output_pdf_path) as pdf:
+        for group in groups_of_interest:
+            print(f"Processing {group} group, type: {type(group)}")
+            metadata_group_df = features_table[[uniq_name, str(group[0])]]
+            merged_depth_df = pd.merge(metadata_group_df, depth_table, how='left', left_on=uniq_name, right_on='SAMPLE_ID')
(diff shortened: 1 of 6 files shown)


efigb:
tested that the branch works in this [seqera run](https://cloud.seqera.io/orgs/bbglabirb/workspaces/Prominent/watch/2EiLrKeMbKXKxe/v2/tasks)

efigb:
Note that to plot the groups they should be parsed in seqera as this:

`features_groups_list: "[ ["BLADDER_LOCATION"], ["SEX"], ["SMOKING_STATUS"] ]" `

When groups are set up in other format it will use only the first group for comparisons (in this case "BLADDER_LOCATION"):

`features_groups_list: "[ ["BLADDER_LOCATION"], ["BLADDER_LOCATION", "SEX"], ["BLADDER_LOCATION", "SEX", "SMOKING_STATUS"] ]"`

Checked [in this run](https://cloud.seqera.io/orgs/bbglabirb/workspaces/Prominent/watch/40JlNlOKqATpXL/v2/tasks)

bin/depth_group_comparison.py
```diff
@@ -0,0 +1,144 @@
...
+
+    return 
+
+
+@click.command()
+@click.option('--table-filename', required=True, type=click.Path(exists=True), help='Input features table file')
+@click.option('--depth-table', required=True, type=click.Path(exists=True), help='Input depth table file')
+@click.option('--separator', required=True, type=click.Choice(['tab', 'comma']), help='Separator used in features table: tab or comma')
+@click.option('--unique-identifier', default=None, type=str, help='Unique identifier column name')
+@click.option('--groups', default=None, type=str, help='List of columns with grouping information')
```
m-huertasp: As this is thought to work with groups, we can modify it so the groups flag is required always


```suggestion
@click.option('--groups', required=True, type=str, help='List of columns with grouping information')
```

bin/depth_group_comparison.py
```diff
@@ -0,0 +1,144 @@
...
+@click.option('--custom-genes', required=False, type=str, help='Comma separated list of custom genes')
+@click.option('--output_prefix', type=str, required=True, help='Prefix for output files')
```
m-huertasp: We discussed about removing the --output_prefix because it is not really needed.

bin/depth_group_comparison.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # groups may contain lists of lists, but all formatted into a string
+    groups_of_interest_init = [group.strip().strip(",").split(",") for group in groups.replace("[", ";;;").replace("]", "").split(";;;")] if groups else []
+
+    groups_of_interest = []
+    for comparison in groups_of_interest_init:
+        comparison_group_clean = [item.strip() for item in comparison]
+        comparison_group = [item for item in comparison_group_clean if item != '']
+        if len(comparison_group) > 0:
+            groups_of_interest.append(comparison_group)
```
m-huertasp: We have discussed that the groups_of_interest_init cannot be modified as it could be used in other parts of the pipeline. The groups_of_interest_init has the following format:
```python
'[ ["BLADDER_LOCATION"], ["BLADDER_LOCATION", "SEX"], ["BLADDER_LOCATION", "SEX", "SMOKING_STATUS"] ]'
```
And we want a plain list with unique items. I think a good strategy could be:

```suggestion
groups_of_interest =  list(dict.fromkeys(item.strip() for sublist in groups_of_interest_init for item in sublist if item != ''))
```

bin/depth_group_comparison.py
```diff
@@ -0,0 +1,144 @@
...
+        panel_genes = sorted(set(custom_gene_list) & set(panel_genes))
+    
+    else:
+        print(f'No custom genes provided, plotting all genes in the panel: {panel_genes}')
+
+    output_name = f"{output_prefix}.plot_depth_per_group.pdf"
+
+    # groups may contain lists of lists, but all formatted into a string
+    groups_of_interest_init = [group.strip().strip(",").split(",") for group in groups.replace("[", ";;;").replace("]", "").split(";;;")] if groups else []
+
```
m-huertasp: Could be changed (if you want) to:

```suggestion
import ast

    groups_of_interest_init = ast.literal_eval(groups) if groups else []
```

bin/depth_group_comparison.py
```diff
@@ -0,0 +1,144 @@
...
+    panel_genes = sorted(set(depth_table['GENE'].unique()))
+    if custom_genes:
+        print(f'Custom genes provided, plotting custom genes only: {custom_genes}')
+        custom_gene_list = [g.strip() for g in custom_genes.split(",")]
+        panel_genes = sorted(set(custom_gene_list) & set(panel_genes))
+    
+    else:
+        print(f'No custom genes provided, plotting all genes in the panel: {panel_genes}')
+
+    output_name = f"{output_prefix}.plot_depth_per_group.pdf"
```
m-huertasp: If you remove the prefix, we should remove it from here as well. In case you want to generate a pdf for each group, you could move the `output_name` to the loop through groups below

modules/local/analyzedepths/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    
+
+
+    output:
+    // the main outputs will be the PDFs
+    path("*.plot_depth_per_group.pdf")                           , emit: plots_per_gene_per_group
+
+    script:
+    // Use meta.id to ensure each sample gets a unique folder/file name
+    def output_prefix = "depth_group_comparison"
```
m-huertasp: We can remove this `output_prefix` as well!

modules/local/analyzedepths/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    depth_group_comparison.py \\
+                --table-filename ${features_table} \\
+                --depth-table ${average_depth_gene_sample} \\
+                --separator ${separator} \\
+                ${unique_identifier} \\
+                ${custom_groups} \\
+                ${custom_genes} \\
+                --output_prefix ${output_prefix}
```
m-huertasp: And this part here as well

workflows/deepcsa.nf
```diff
@@ -227,6 +229,17 @@ workflow DEEPCSA{
...
+    // define it as a module very similar to the table2group one
+    ANALYZEDEPTHSGROUPS(features_table, PLOTDEPTHSEXONSCONS.out.average_depth_gene_sample)
+
+    // Load group keys from JSON file in 'groups' channel
+    TABLE2GROUP.out.json_groups.map { json_path ->
+        def json = file(json_path).text
+        groovy.json.JsonSlurper.newInstance().parseText(json).keySet()
+    }.flatten().unique()
+    .set { group_keys_ch } // this is a channel that contains only the group names as elements of the channel
+
```
m-huertasp: I think you don't really need to add this in here, no? It is already done above.

workflows/deepcsa.nf
```diff
@@ -227,6 +229,17 @@ workflow DEEPCSA{
     }
     PLOTDEPTHSEXONSCONS(ANNOTATEDEPTHS.out.all_samples_depths, CREATEPANELS.out.exons_consensus_bed, CREATEPANELS.out.exons_consensus_panel)
 
+    // define it as a module very similar to the table2group one
+    ANALYZEDEPTHSGROUPS(features_table, PLOTDEPTHSEXONSCONS.out.average_depth_gene_sample)
```
m-huertasp: We discussed about doing this process only if there are groups defined. We could do something like:

```suggestion
if (params.features_groups_list) {
    ANALYZEDEPTHSGROUPS(features_table, PLOTDEPTHSEXONSCONS.out.average_depth_gene_sample)
}
```

(I think)

m-huertasp (commented):
Hola!! I think this is quite a nice job! As we talked about, there are a few things that might need to be changed. I have added some comments so it is easier to check the things we discussed.

Good job!

bin/depth_group_comparison.py
efigb: Totally agree to remove it, it does nothing meaningful here

bin/depth_group_comparison.py
```diff
@@ -0,0 +1,139 @@
...
+    pdf.savefig()
+    plt.close()
+    plt.show()
+
+    return 
+
+
+@click.command()
+@click.option('--table-filename', required=True, type=click.Path(exists=True), help='Input features table file')
+@click.option('--depth-table', required=True, type=click.Path(exists=True), help='Input depth table file')
```
efigb: @FerriolCalvet I couldn't find an output file from consensus exons depth that contains the "ALL_GENES" value per sample, so that I could use it to plot the all genes value per sample. These are the other files from the output, I used average_depth_gene_sample:
```
    average_depth_sample        = DEPTHSSUMMARY.out.average_per_sample
    average_depth_gene          = DEPTHSSUMMARY.out.average_per_gene
    average_depth_gene_sample   = DEPTHSSUMMARY.out.average_per_gene_sample
```

So the script uses the average_depth_gene_sample to plot the 'all genes' value per group of samples, and this is why the 'all genes' plots per group show more spots than the by gene per group plots. The way it handles it could be error prone so I would try to fix this with an external file.

conf/modules.config
```diff
@@ -336,6 +336,21 @@ process {
...
+    withName: "ANALYZEDEPTHSGROUPS" {
+        // these params.features are defined in nextflow.config
+        ext.unique_identifier           = params.features_unique_identifier
+        ext.features_groups             = params.features_groups_list
+        ext.separator                   = params.features_table_separator
+        ext.custom_genes                = params.features_genes_list
+        // define the list of custom genes here
+        // you will have to add an extra parameters to the pipeline (nextflow.config) and then handle it here
+        publishDir = [
+            path: { "${params.outdir}/plots/depths_summary" },
```
efigb: @FerriolCalvet consider whether it makes sense to create a subdirectory for depth-per-groups-plots, since the number of pdf files generated will match the number of groups defined by the user

FerriolCalvet: sure, feel free to create it, you just need to define here the name that you would like it to have

bin/depth_group_comparison.py
FerriolCalvet: there is no such ALL_GENES scenario, the information in the `average_depth_sample` is like what you would expect to have for ALL_GENES.
Maybe you could:
1. also take this file
2. define a GENE column with the value ALL_GENES 
3. then concatenate this table to the table that you already have
and this way you would have a correct ALL_GENES value that you could plot with the rest of plots

bin/depth_group_comparison.py
```diff
@@ -93,10 +101,36 @@ def main(table_filename, depth_table, unique_identifier, separator, groups, cust
...
+    # Merge data with metadata table to get the group information for each sample
+    merged_depth_df = pd.merge(features_table, depth_genes_samples, how='left', left_on=uniq_name, right_on='SAMPLE_ID')
+    print('Merged depth and metadata table:')
+    print(merged_depth_df.head())
+    print('Length of merged table:', len(merged_depth_df))
+
     # Process groups
-    groups_of_interest_init = ast.literal_eval(groups) if groups else []
-    groups_of_interest = []
-    groups_of_interest =  list(dict.fromkeys(item.strip() for sublist in groups_of_interest_init for item in sublist if item != ''))
```
efigb: there was a bug in this code that I had to replace to handle different type of groups:

  `["BLADDER_LOCATION"], ["BLADDER_LOCATION", "SEX"], ["BLADDER_LOCATION", "SEX", "SMOKING_STATUS"] ]'`

and

`"[ ["Sample_Group"], ["cancer"], ["Age_onset"], ["Cancer_age_group"] , ["Bacterial_Signatures_identified"]]"`

bin/depth_group_comparison.py
```diff
@@ -93,10 +101,36 @@ def main(table_filename, depth_table, unique_identifier, separator, groups, cust
     else:
         print(f'No custom genes provided, plotting all genes in the panel: {panel_genes}')
 
+
+    # Process depth per sample to add the 'ALL_GENES' depth value per sample
+    print('Processing per sample depth table to add the ALL_GENES depth column: ')
+    depth_per_sample['GENE'] = 'ALL_GENES'
```
efigb, line 107: @FerriolCalvet if user does not define the genes in the panel as custom genes, with the way we handle it now to calculate ALL_GENES depth, it will display all depths from both on target and off target genes in the panel right? I mean the default would not be the panel genes but including off targets...

conf/modules.config
```diff
@@ -345,7 +345,7 @@ process {
         // define the list of custom genes here
         // you will have to add an extra parameters to the pipeline (nextflow.config) and then handle it here
         publishDir = [
-            path: { "${params.outdir}/plots/depths_summary" }, 
+            path: { "${params.outdir}/plots/depths_summary/plots_per_group" },
```
efigb, line 348: now added as subdirectory specific for groups

bin/depth_group_comparison.py
FerriolCalvet: it depends, since the input to this step comes from the consensus_exons file, if the user defines some selected genes for building this panel then the ALL_GENES would not contain off targets. But would contain all genes not only those where the user might request to get more detailed plotting.

I think that the way it is being done now makes sense.

bin/depth_group_comparison.py
FerriolCalvet: I will check it properly once I review the PR, but I am not sure if we might need an additional parameter to define the groups, if the groups come from the same parameter as the group comparisons for the other metrics, this might not be ideal...

We can discuss once I have more deeply reviewed everything.

efigb:
tested branch in pipeline with seqera and works

PR #425, opened 2026-02-26
add decomposition of HDP with SigProfilerAssignment and minor fixes

### LINKED ISSUE #298: Run SigProfilerAssignment with the pipeline output of HDP
## We still need to agree whether this is necessary or not, but if we want to implement it, this would be the way to go.

- [ ] Set an output channel for HDP signatures
- [ ] Pass this output to a new call to SigProfilerAssignment and use this as a reference signatures
- [ ] Make sure you set a proper definition on where the output should be stored

@rochamorro1 
these could be the steps to follow I am adding it just for reference in case we want to solve it and so that we do not forget

### LINKED ISSUE #364: Mutational profile stability

### LINKED ISSUE #409: Reduce duplicated mutation subsets
In multiple subworkflows one of the first steps is using tabix to subset the mutations to the regions of interest, usually the same exons consensus panel, we could take an approach similar to the one for the depths that reduces the mutations only to the regions of interest

FerriolCalvet:
## Main update: extracted signatures decomposition
The main update is to do the decomposition of the mutational signatures extracted by HDP using SigProfilerAssignment.
This results in a much cleaner assignment of mutational signatures to samples. See example below:

SigProfilerAssignment directly:
image

Results of SigProfilerAssignment decomposition using the signatures extracted by HDP:
image

### Minor fixes
- Remove all mentions to hg19 and mm10 these are not supported anymore
- Remove some duplicated mutation subsets for exonic mutations

### DIFF
--- a/bin/concat_sigprot_matrices.py
+++ b/bin/concat_sigprot_matrices.py
@@ -41,6 +41,13 @@ def concat_sigprot_matrices(filename_of_matrices, samples_json_file, type_of_pro
 
     # Save the groups matrix if it exists
     if groups_matrix is not None and groups_matrix.shape[0] > 0:
+
+        # identify samples without mutations
+        zero_rows = groups_matrix.index[groups_matrix.sum(axis=1) == 0]
+        if len(zero_rows) > 0:
+            print(f"Excluding samples without mutations: {list(zero_rows)}")
+            groups_matrix = groups_matrix.drop(zero_rows)
+
         groups_matrix.reset_index().to_csv(f"groups_matrix.{type_of_profile}.sp.tsv", sep='\t', header=True, index=False)
         groups_matrix.round().astype(int).reset_index().to_csv(f"groups_matrix.{type_of_profile}.sp.round.tsv", sep='\t', header=True, index=False)
 
@@ -55,6 +62,13 @@ def concat_sigprot_matrices(filename_of_matrices, samples_json_file, type_of_pro
 
     # Save the samples matrix if it exists
     if samples_only_matrix is not None and samples_only_matrix.shape[0] > 0:
+
+        # identify samples without mutations
+        zero_rows = samples_only_matrix.index[samples_only_matrix.sum(axis=1) == 0]
+        if len(zero_rows) > 0:
+            print(f"Excluding samples without mutations: {list(zero_rows)}")
+            samples_only_matrix = samples_only_matrix.drop(zero_rows)
+
         samples_only_matrix.reset_index().to_csv(f"samples_matrix.{type_of_profile}.sp.tsv", sep='\t', header=True, index=False)
         samples_only_matrix.round().astype(int).reset_index().to_csv(f"samples_matrix.{type_of_profile}.sp.round.tsv", sep='\t', header=True, index=False)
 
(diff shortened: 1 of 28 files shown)


efigb:
Other tests in other cohorts like colon:

With all 3.4 COSMIC signatures (sorry about the quality):

image

With reffiting of CRC signatures:

image

bin/mut_profile.py
```diff
@@ -134,6 +201,13 @@ def compute_mutation_profile(sample_name, mutation_matrix_file, trinucleotide_co
     contextmut_depth = pd.DataFrame({sample_name : trinuc_per_contextmuts, "CONTEXT_MUT": contexts_formatted})
     contextmut_depth = contextmut_depth.set_index("CONTEXT_MUT")
 
+    # compute profile stability and store the outputs in a file
+    stability_output_dict = profile_stability(mutation_matrix[sample_name], contextmut_depth[sample_name])
+    stability_outputs = pd.DataFrame(stability_output_dict.items()).set_index(0).T
+    stability_outputs[["SAMPLE_ID", "mode"]] = sample_name.split(("."))
```
m-huertasp: This is not important at all because it is not going to change anything, pero bueno


```suggestion
    stability_outputs[["SAMPLE_ID", "mode"]] = sample_name.split(".")
```

bin/concat_sigprot_matrices.py
```diff
@@ -41,8 +49,11 @@ def concat_sigprot_matrices(filename_of_matrices, samples_json_file, type_of_pro
 
     # Save the groups matrix if it exists
     if groups_matrix is not None and groups_matrix.shape[0] > 0:
+        groups_matrix = remove_zero_mutation_samples(groups_matrix)
+
         groups_matrix.reset_index().to_csv(f"groups_matrix.{type_of_profile}.sp.tsv", sep='\t', header=True, index=False)
```
m-huertasp, line 54: I don't know if I've checked everything, but the non-rounded version is not really used, no? For me, it is a bit confusing. If it is not used I would remove it, but I don't really have strong opinions.

m-huertasp (commented):
I think everything should be good but I had a problem while testing the PR in the SIGPROFILERASSIGNMENT step. The error is:

```bash
Something is wrong with the format of signature database, Pass a text file of signatures in the format of COSMIC sig database
```
I checked and the path in the configuration (general_files_IRB) is the last version of the cosmic reference signatures in that folder. I don't know If there is something that I should be doing differently or if the file is damaged. I don't have permissions to read it so I can't really take a look.

I did a quick review of the code and left two small and not really important comments.

bin/concat_sigprot_matrices.py
FerriolCalvet: It is true that it is not used, but at least being part of the output allows the user to check which is the original values of each sample & mutation context

FerriolCalvet:
I will double check this, because it is true that I have seen this error before so I will test it again and check.
Thanks for the review!

FerriolCalvet:
I tested it and locally it worked for me with both the 3.5 and 3.4 I guess it was a permissions issue and now I updated the permissions of the 3.5 version so I guess it should work for you as well.

I will merge it now.

m-huertasp:
Yess, working fine now! Thanks!

FerriolCalvet:
thank you for testing again Marta!! 🙏🏻

PR #433, opened 2026-03-18
disable automatic filling of 0s in profile

FerriolCalvet:
- Add the option to smooth the mutational profile when having less than 200 mutations in the sample (this is hardcoded since we conducted entropy tests to define the threshold).
- Fix the flagging of omegas.
- Added the computation of proportion of SNP sites mutated to inform contamination in samples independent from the samples being analyzed in the same batch/cohort.

### DIFF
--- a/bin/mut_profile.py
+++ b/bin/mut_profile.py
@@ -10,6 +10,17 @@ from utils import contexts_formatted, contexts_formatted_sigprofiler
 from utils_plot import plot_profile
 from read_utils import custom_na_values
 
+def bayesian_update(observed_counts, prior_profile, limit=200):
+    
+    N = np.sum(observed_counts)
+    
+    if N >= limit:
+        return observed_counts / N
+    else:
+        posterior_profile = (N - limit) * prior_profile + observed_counts
+        posterior_profile /= np.sum(posterior_profile)
+        return posterior_profile
+
 
 def compute_mutation_matrix(sample_name, mutations_file, mutation_matrix, method, pseudocount,
                             sigprofiler, per_sample):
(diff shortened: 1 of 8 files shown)


bin/mut_profile.py
```diff
@@ -271,14 +286,36 @@ def compute_mutation_profile(sample_name, mutation_matrix_file, trinucleotide_co
...
+            
+            upd_mutation_matrix_wgs = bayesian_update(profile_trinuc_clean, prior_profile, minimum_mutations).reset_index()
+            total_mutations = max(minimum_mutations, total_mutations)
+
+            # we have updated the mutation counts at the WGS level, we should now revert this update
+            # for the specific sample's trinucleotide content and depth
+
+            upd_mutation_matrix_wgs["CONTEXT"] = upd_mutation_matrix_wgs["CONTEXT_MUT"].apply( lambda x : x[:3])
+            profile_trinuc_merge = upd_mutation_matrix_wgs.merge(ref_trinuc32, on = "CONTEXT")
+            profile_trinuc_merge["MUT_PROBABILITY"] = profile_trinuc_merge[sample_name] / profile_trinuc_merge["COUNT"]
```
koszulordie: The definition of MUT_PROBABILITY is correct to represent the relative mutability per sequenced site given the triplet context, but it is not as intuitive, so we can consider two actions here:
1) Adding a comment explicitly defining what this "probability" represents
2) If this is intended to represent the relative mutability per sequenced tri-nucleotide context, which will be later scaled to match an (absolute) expectation per site sequenced, for subsequent calculations it will be no harm to provide this profile already normalized, meaning that the channels sum up to 1.

koszulordie (commented):
I focused on the mut_profile.py script and as far as I can tell everything looks correct. I commented on the definition of MUT_PROBABILITY in line 302, we can add a comment here for clarity, as this step is critical. I also suggest to provide this profile in normalized form so that the scale, although arbitrary, always displays the same range of values.

bin/mut_profile.py
FerriolCalvet: discussed and agreed!

FerriolCalvet:
we agreed this branch is ready to be merged, only pending to be fully tested with the update that removes the pseudocount of relative mutability per channel

once we merge #390

PR #438, opened 2026-03-27
TESTS: testing with small bladder set

### LINKED ISSUE #435: Use minimum average depth parameter when running with custom depths

m-huertasp:
This PR is intended to make sure that the nf-tests work to test the pipeline and to ensure the test data used is prepared to be shared with users. 

Added `FILTERDEPTHS` module to allow depth-based filtering on input tables (matching our standard computation logic).
**Key changes:**
### 1. Test Infrastructure and Documentation Improvements

- Updated `tests/README.md` to document cluster-based testing, test configuration, output directories, and how to adapt the suite for other environments. Added instructions for new test cases, cleaning outputs, and debugging. [[1]](diffhunk://#diff-dacac2ebf9792f0d23c0f922a744486ded01901957d5281290925acd89cf83acL3-R28) [[2]](diffhunk://#diff-dacac2ebf9792f0d23c0f922a744486ded01901957d5281290925acd89cf83acR38-R98) [[3]](diffhunk://#diff-dacac2ebf9792f0d23c0f922a744486ded01901957d5281290925acd89cf83acL91-R164) [[4]](diffhunk://#diff-dacac2ebf9792f0d23c0f922a744486ded01901957d5281290925acd89cf83acL110-R183)
- Updated `nf-test.config` to use a scratch work directory, ignore nf-core modules/subworkflows, and load the `nft-utils` plugin.

### 2. Expanded and Refactored Test Suite

- Refactored and expanded `tests/deepcsa.nf.test` to include:
  - A standard functionality test with improved assertions and updated output paths.
  - An omega analysis test with checks for output structure, sample completeness, and directory layout.

### 3. Resource and Parameter Updates

- Increased resource limits for test processes in `conf/test.config` to better match cluster capabilities (4 CPUs, 20 GB RAM, 15 min).

### 4. New Depth Filtering Module and Workflow Integration

- Added a new process `FILTERDEPTHS` in `modules/local/filterdepths/main.nf` that filters depth tables based on a configurable minimum mean depth threshold.
- Integrated the `FILTERDEPTHS` module into the `DEPTH_ANALYSIS` workflow in `subworkflows/local/depthanalysis/main.nf`, so that when `use_custom_minimum_depth` is set, depth tables are filtered accordingly. [[1]](diffhunk://#diff-d55ce23fd23e7d8b133f559e9ac9dc0d7ba74fb63f6c3b3ea04f4e429783a66dR2) [[2]](diffhunk://#diff-d55ce23fd23e7d8b133f559e9ac9dc0d7ba74fb63f6c3b3ea04f4e429783a66dL13-R21)
- Registered the `FILTERDEPTHS` process in the Nextflow configuration (`conf/modules.config`) with parameter wiring for `minimum_depth`.

These changes collectively add a new filtering capability, improve pipeline robustness, and ensure the test suite is comprehensive and portable across different SLURM cluster environments.

closes #435


tests/test_data/input.csv
```diff
@@ -1,16 +1,4 @@
...
-B6,/data/bbg/datasets/pipelines/deepCSA/test_datasets/deepUMIcaller_outputs/mutations_vcf/B6.duplex.filtered.vcf,/data/bbg/datasets/pipelines/deepCSA/test_datasets/deepUMIcaller_outputs/sortbamduplexcons/B6.sorted.bam
-B7,/data/bbg/datasets/pipelines/deepCSA/test_datasets/deepUMIcaller_outputs/mutations_vcf/B7.duplex.filtered.vcf,/data/bbg/datasets/pipelines/deepCSA/test_datasets/deepUMIcaller_outputs/sortbamduplexcons/B7.sorted.bam
-B8,/data/bbg/datasets/pipelines/deepCSA/test_datasets/deepUMIcaller_outputs/mutations_vcf/B8.duplex.filtered.vcf,/data/bbg/datasets/pipelines/deepCSA/test_datasets/deepUMIcaller_outputs/sortbamduplexcons/B8.sorted.bam
-B9,/data/bbg/datasets/pipelines/deepCSA/test_datasets/deepUMIcaller_outputs/mutations_vcf/B9.duplex.filtered.vcf,/data/bbg/datasets/pipelines/deepCSA/test_datasets/deepUMIcaller_outputs/sortbamduplexcons/B9.sorted.bam
+sample,vcf
+P19_0002_BDO_01,path/to/P19_0002_BDO_01.vcf
+P19_0002_BTR_01,path/to/P19_0002_BTR_01.vcf
+P19_0003_BDO_01,path/to/P19_0003_BDO_01.vcf
```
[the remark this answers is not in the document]
m-huertasp: I need to change the README. The vcf field is needed for the samplesheet to pass the checks but not really used. The inputs used are in `tests/nextflow.config` : the MAF file for all the samples and the depth table.

nf-test.config
```diff
@@ -1,9 +1,20 @@
 config {
-
     testsDir "tests"
-    workDir ".nf-test"
+
+    // TODO: Change this to read for an environment variable instead of hardcoding it here?
+    // Working files from the test
+    workDir "/scratch/bbg/work/deepCSA/nf-tests"
```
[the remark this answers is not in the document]
m-huertasp: Thinking about doing this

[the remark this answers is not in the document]
FerriolCalvet: looks good to me, it will make things better than having the work hidden in the same director where things are run

tests/test_data/input.csv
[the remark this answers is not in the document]
FerriolCalvet: I think that in the scenario where the MAF and the depths are provided the input.csv is used to list the samples that are available for analysis so if this is not properly coordinated it would fail.

Regarding the BAMs, yes you can ignore the suggestion I guess

m-huertasp:
## UPDATE
This is a comment so we know what was done here when coming back from holidays:
- Removed all hardcoded cluster paths from the test setup so the tests are portable across environments without manual file editing. **Missing**: Should we do the same with the container? Currently the path to IRB cluster is hardcoded.
- **Test data**: MAF and depths files are now fetched at runtime directly from bbglab/DeepClone_protocol via raw GitHub URLs (set in the when { params } block of [deepcsa.nf.test](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)). validation.ignoreParams added to [nextflow.config](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) so nf-schema skips file-existence checks for those two params.
- **Work directory**: workDir in [nf-test.config](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) now reads the NFT_TEST_WORKDIR environment variable (falling back to [.nf-test](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in the repo root). Added info in the README about this and the environment variant to set up.
- **Output directory**: Per-test outdir now uses nf-test's built-in $outputDir variable (auto-derived from workDir).
- **README**: Updated to reflect all of the above, including setup instructions for a new cluster environment.

As the data is in DeepClone, the tests will not work until the repo is public.

bin/utils_filter.py
```diff
@@ -143,7 +143,7 @@ def extract_flagged_regions_bed(maf_df: pd.DataFrame, name: str, FILTERS: list[s
         Output coordinates are 0-based with half-open intervals [start, end).
     """
     # List of filter columns you want to check for
-    filter_columns = [f"FILTER.{f}" for f in FILTERS if f in ','.join(list(maf_df.columns))]
```
FerriolCalvet, line 146: this line was a bit weird, thanks for the fix

conf/test.config
```diff
@@ -16,9 +16,8 @@ params {
 }
 
 
-
 process {
     errorStrategy = 'retry'
     maxRetries = 2
-    resourceLimits = [ cpus: 1, memory: 10.GB, time: 10.m ]
+    resourceLimits = [ cpus: 4, memory: 20.GB, time: 15.m ]
```
FerriolCalvet: if the tests take very long we can think of incresing CPUs here, I don't think it should be a problem to raise it to 8 at least, even a laptop could handle it.

tests/deepcsa.nf.test.snap
FerriolCalvet, line 1: did you check if the changes to the files were big? or just minor changes on decimal points and so on?
more out of curiosity than nothing else

tests/nextflow.config
```diff
@@ -4,81 +4,40 @@
...
-    space = 200
-    cpiter = 3
-    iterations = 15
-    cosine_sim_threshold = 0.9
-    max_iter_em = 1000
-    em_frac_threshold = 0.1
+    config_profile_name        = 'Test profile'
+    config_profile_description = 'Test dataset to check pipeline function'
+
+    use_custom_minimum_depth   = 10   // main default: 0
```
FerriolCalvet: we should increase it to 20 or 30 also the default (dev default should be the one to use)

it is pointless to use processing time for regions so poorly covered

tests/README.md
```diff
@@ -1,24 +1,31 @@
...
-- `tests/test_data/`: Small input dataset used by the tests
-- Output (created by tests):
-  - `${projectDir}/tests_results_normal`
-  - `${projectDir}/tests_results_omega`
+- `tests/nextflow.config`: Test-specific Nextflow config (SLURM, Singularity, cluster paths, param overrides)
+- `tests/test_data/`: Local committed input files used by the tests (`input.csv`, `test_mutations.maf`). The MAF and depths files for the main tests are fetched at runtime directly from [bbglab/DeepClone_protocol](https://github.com/bbglab/DeepClone_protocol) — no download step required.
 
 ### Running Tests
+
+> **Prerequisite:** Tests require a SLURM cluster with Singularity available. All jobs are submitted to SLURM automatically — do not run from a local machine, as resource limits will not be met. If you are not on the IRB cluster, follow [Configuring for usercase](#configuring-for-usercase) first.
```
FerriolCalvet, line 27: 👍🏻

tests/README.md
FerriolCalvet, line 1: I know this would not be for general use, but it would be great to have an example of which are the exact steps that you follow, I guess you may be indicating -outputDir or other flags in the command-line instructions so this way we have it explicitly defined here and there is no doubt on how to handle the tests efficiently.
(maybe a small section at the end would be the cleanest and simplest to not be misleading)

Maybe I am completely off and this should not be here eh!

FerriolCalvet (approved):
Looks good Marta thanks for the cleaning and improvement in documentation of all the testing machinery.
I see that all the parameters that are the same as the default ones stay only in the main nextflow.config file, I guess this means that as long as we do not change the defaults, the tests should be fine, and that whenever we update the defaults we should update the snapshots right?
Maybe we should indicate this somewhere.

I think this is ready to merge and it would be great to make the DeepClone repo public ASAP so that we can test it directly with the paths referencing the repo.

I am adding @migrau  in case he wants to take a look at it, but in any case we will try to merge it this week, and then if needed apply patches later.

Also as discussed offline in the past with both of you, probably some steps would benefit more from having unit tests rather than full pipeline tests since it is to wide and intertwined that it is complicated to know what is being tested in each execution unless you know exactly how everything works.

FerriolCalvet (approved):

tests/deepcsa.nf.test.snap
m-huertasp: Now that you've commented I've realized that the snapshot should be updated... As I changed the input to only 3 bladder samples I didn't check but it could be interesting..

tests/nextflow.config
m-huertasp: Should I change it or it is better to discuss this in a pipelines meeting?

FerriolCalvet: change it to 30

m-huertasp:
Missing:
- Update snapshots once DeepClone is public
- Check if there are some instructions missing on how to run the tests

m-huertasp:
Already checked that tests PASS and snapshot updated.

PR #444, opened 2026-04-11
add dynamic RefCDS definition for dNdScv

### LINKED ISSUE #240: Build RefCDS object. maybe from consensus exons panel?

FerriolCalvet:


FerriolCalvet:
we should make sure that the transcripts selected in our consensus panel and the ones downloaded from the Ensembl Biomart query are the same otherwise this could lead to several artificial splice sites that are solvable

bin/biomart_query.R
```diff
@@ -0,0 +1,114 @@
...
+    '<Attribute name = "cds_end" />',
+    '<Attribute name = "cds_length" />',
+    '<Attribute name = "strand" />',
+    '<Attribute name = "ensembl_transcript_id" />',
+    '<Attribute name = "exon_chrom_start" />',
+    '<Attribute name = "exon_chrom_end" />',
+    '</Dataset>',
+    '</Query>'
+  )
+  return(query)
```
FedericaBrando: This works, but if you are using R by default then maybe it would be worth to use biomart R package directly, maintained by ensembl people directly. 

https://bioconductor.org/packages//release/bioc/manuals/biomaRt/man/biomaRt.pdf

modules/local/dnds/filterbiomart/main.nf
```diff
@@ -0,0 +1,45 @@
+process FILTER_BIOMART {
```
FedericaBrando: nitpick: The name is quite broad and not very self-explanatory of what the process is actually doing, I would recommend something like: "BUILD_PANEL_BIOMART", "ADAPT_PANEL_REFCDS", something that actually explains that the script adpats the region to be compatible with the building of the RefCDS

modules/local/dnds/adaptpanelrefcds/main.nf
```diff
@@ -0,0 +1,45 @@
...
+                        --genes ${biomart} \\
+                        --output custom_filtered_biomart.tsv \\
+                        --verbose \\
+                        -s splice_sites.tsv
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        Ensembl BioMart: v111
+    END_VERSIONS
+    """
```
FedericaBrando, line 31: You can run `dNdScv_panel_prep.py --version` and it outputs the version

FedericaBrando (commented):
Overall it looks good, I think I will steal the nextflow process to implement the build of the custom RefCDS in IntOGen ahah

I added some comment, regarding biomart I would not reinvent the wheel, the query the way it is implemented now is a bit fragile, I would use ensembl packages, this is good for future development (raw query is harder to debug, understand, compared to a documented package) and for robustness (better error handling). 

Another nitpick is about naming the process "FILTER_BIOMART" I think it doesn't fully explain what the process is doing, I would use the work ADAPT or something that let the user understand that the step is required for the next one, not so much linked to the previous (not sure if I am clear - I can extend the explanation if needed).

Anyway, overall looks good! I will leave the review as comment, up to you if you want to address the above mentioned comment or merge! Great job 🚀

FerriolCalvet:
thanks for the comments Fede! I will address them and then merge it since it is currently not fully functional for bigger panels, and passing this problem onto Ensembl functions sounds better

FerriolCalvet:
@FedericaBrando I have worked a bit on this, and more or less it is functional but there are two main caveats that make me doubt if this should be merged or we should undergo some bigger reimplementation.
- The querying of Ensembl Biomart via their tool can work well, but does not always work and in cases of big batches such as the exome panel it easily brakes at some point in while querying some batch...
- I am querying the latest Ensembl verison since the archive of jan2024 was working very poorly and I rarely got the process to succeed. This means that the annotation is mismatched and there could be issues on top of us (PROMINENT) not using the MANE transcripts, that is also something that we should fix.

Possible alternatives:
- provide a gtf file with the annotation of Ensembl v111 and derive all the metrics from there
- [insert your ideas]

Let me know what you think, and let's see if we can get this solved soon... 🤞🏻

PR #449, opened 2026-05-04
add dn/ds proxy

FerriolCalvet:
This branch contains: 
- minimal updates in the computations of the profile-adjusted mutation density
- the addition of dN/dS proxy computation that is obtained by dividing the adjusted mutation density of missense or truncating mutations by the synonymous mutation density value either of the sample or of the whole cohort.
- plotting of interindividual variability in the adjusted mutation density


FerriolCalvet (commented):
all checked and looks good

PR #469, opened 2026-06-02
refactor: uniformize somatic-variant definition across bin scripts

### LINKED ISSUE #418: uniformize the definition of somatic variants
In `filter_cohort.py` the definition of somatic variants is done by using "VAF", the line is the following:

``` python
maf_df_f_somatic = maf_df.loc[maf_df["VAF"] <= somatic_vaf_boundary][["MUT_ID","SAMPLE_ID", "PMEAN", "PSTD"]].reset_index(drop = True)
```

In other places the definition of somatic is made filtering also for the "vd_VAF" and "VAF_AM" columns. We should uniformize this definition and make sure that it is the same everywhere.

m-huertasp:
Small PR.

## What this does
Determines single definition of a *somatic* variant and applies it
consistently across the `bin/` scripts. Previously the somatic/germline split was
re-implemented inline in several places with divergent forms (single-column `VAF`
in `filter_cohort.py`, three-column elsewhere, plus hardcoded thresholds in
`check_contamination.py`), which made the somatic and germline sets non-symmetric.
The definition now lives in two shared helpers and every call site uses them. Closes #418.

## Changes
- Add `somatic_mask` / `germline_mask` to `bin/utils_filter.py` — the canonical all-3-column
  rule (`VAF`, `vd_VAF`, `VAF_AM` vs a caller-supplied threshold), with NumPy docstrings.
- Refactor `bin/filter_cohort.py` to use the helpers (somatic flagging goes single-column → all-3-column).
- Refactor `bin/check_contamination.py`: germline/SNP predicates use the helpers; the hardcoded
  `0.2` threshold is replaced by a new `--somatic-vaf-boundary` CLI option fed from
  `params.germline_threshold` (wired via `modules/local/contamination/main.nf` + `conf/modules.config`);
  NumPy docstrings added to all public functions.
- Add `bin/test/test_utils_filter.py` — 46 unit tests covering every public function in `utils_filter.py`.

## What to review
- **Behaviour changes:** `filter_cohort` somatic flagging is now all-3-column; the
  between-samples contamination threshold moves `0.2 → 0.3` (default `germline_threshold`).

## Testing
`pytest bin/test/test_utils_filter.py` → 46 passed + ruff clean on all touched files.
The full `bin/test/` suite has 3 pre-existing failures + 1 collection error in unrelated files
(`test_plot_selectionsideplots.py`, `test_check_samplesheet.py`) that are untouched by this PR.

### DIFF
--- a/bin/check_contamination.py
+++ b/bin/check_contamination.py
@@ -454,7 +454,7 @@ def contamination_detection_in_snps(maf):
     somatic_snp_positions_maf = snp_positions_maf.loc[
         somatic_mask(snp_positions_maf, contamination_vaf_threshold)
     ].reset_index(drop=True)
-    germline_snp_positions_maf = snp_positions_maf[
+    germline_snp_positions_maf = snp_positions_maf.loc[
         germline_mask(snp_positions_maf, contamination_vaf_threshold)
     ].reset_index(drop=True)
 
--- a/bin/check_contamination.py
+++ b/bin/check_contamination.py
@@ -6,10 +6,11 @@ Script for checking contamination between samples.
 
 
 import click
-import pandas as pd
 import matplotlib.pyplot as plt
+import pandas as pd
 import seaborn as sns
 from read_utils import custom_na_values
+from utils_filter import germline_mask
 
 
 # Assuming somatic_variants and germline_variants are loaded as pandas DataFrames
@@ -45,7 +46,7 @@ def contamination_detection_between_samples(maf_df, somatic_maf_df):
 
     # this is if we were to consider both unique and no-unique variants
     vaf_threshold = 0.2
-    germline_vars_all_samples = maf_df.loc[(maf_df["VAF"] > vaf_threshold) & (maf_df["vd_VAF"] > vaf_threshold) & (maf_df["VAF_AM"] > vaf_threshold),
+    germline_vars_all_samples = maf_df.loc[germline_mask(maf_df, vaf_threshold),
                                             ["SAMPLE_ID", "MUT_ID"]].drop_duplicates()
 
     print(germline_vars_all_samples["MUT_ID"].shape)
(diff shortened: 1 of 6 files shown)


bin/check_contamination.py
FerriolCalvet, line 1: I think that there is something that pisses me off a bit from this script and it is that the sequence of plots in the code is a bit confusing.

Could you give it a try and split the main function into different ones? I can give it a try as well otherwise.

bin/check_contamination.py
```diff
@@ -351,38 +399,64 @@ def contamination_detection_between_samples(maf_df, somatic_maf_df):
...
     # being very restrictive in the VAF to count the occurrences of potentially contaminated mutations
-    somatic_snp_positions_maf = snp_positions_maf[snp_positions_maf["VAF"] < 0.05].reset_index(drop = True)
-    germline_snp_positions_maf = snp_positions_maf[snp_positions_maf["VAF"] >= 0.05].reset_index(drop = True)
+    contamination_vaf_threshold = 0.05
+    somatic_snp_positions_maf = snp_positions_maf.loc[
+        somatic_mask(snp_positions_maf, contamination_vaf_threshold)
+    ].reset_index(drop=True)
+    germline_snp_positions_maf = snp_positions_maf.loc[
+        germline_mask(snp_positions_maf, contamination_vaf_threshold)
+    ].reset_index(drop=True)
```
FerriolCalvet: I am not sure that this has the same behaviour than before since there might be some mutations that stay in between the two

FerriolCalvet: maybe inverting the somatic_mask would be the best way to go?

bin/filter_cohort.py
FerriolCalvet, line 1: lots of spaceees and quotes hahaha

FerriolCalvet (commented):
looks good, there is only this comment on the check_contamination script that if you can address it to increase the clarity that would be great, if you think that it is clear enough I am happy to accept that as well

bin/check_contamination.py
m-huertasp: it does not have the same behaviour than before because we are using the three VAFs and we can loose mutations that are VAF > 0.05 but vd_VAF <0.05. Would it be best to use the previous strategy in this case?

FerriolCalvet: I am not sure that it makes a big difference, but probably yes, it is better to follow the same strategy as before.
I think that my suggestion of using the somatic_mask and then "inverting" the selection should work well and would not be too difficult to apply.

PR #479, opened 2026-07-02
NOT MERGE 478 qc plot on omega vs dndscv values comparison

FerriolCalvet:


subworkflows/local/plotting_qc/main.nf
```diff
@@ -54,14 +54,19 @@ workflow PLOTTING_QC {
 
     APPLYOMEGAQC(all_omegas, PLOTMUTDENSITYQC.out.compiled_flagged.collect())
     
-    PLOTOMEGAVSDNDSCV(all_omegas, dndscv_cv, groups_definition, group_name, APPLYOMEGAQC.out.flagged_cases)
```
efigb: @FerriolCalvet I'm not sure that the pipeline does not proceed due to this code lines

bin/omega_vs_dndscv_qc.py
```diff
@@ -0,0 +1,161 @@
...
+def main(input_omega_file, input_dndscv_file, output_dir, flagged_genes_omega): 
+    
+    # Read tables
+    omega_df = pd.read_table(input_omega_file, sep='\t')
+    dndscv_df = pd.read_table(input_dndscv_file, sep='\t')
+    flagged_genes_df = pd.read_table(flagged_genes_omega, sep='\t')
+
+    # Extract sample groups
+    sample_groups = flagged_genes_df['cohort'].unique().tolist()
+
```
m-huertasp: I understand that the idea of this line is to get the name of the sample groups that you'll use later to generate the plot, right? Is the idea to plot only groups in which there is at least one flagged gene or you want to plot them all? Because by getting the groups from the `flagged_genes_df` you are getting only those with at least one flagged gene.

If you want to plot all, maybe you can get the groups from the `omega_df` or the `dndscv_df`??

bin/omega_vs_dndscv_qc.py
```diff
@@ -0,0 +1,161 @@
...
+                corr, pval = pearsonr(impact_subset_reg['dnds_omega'], impact_subset_reg['dnds_dndscv'])
+                
+                ax.set_title(f'{impact.capitalize()} mutations\nPearson R: {corr:.3f} (p={pval:.3e})')
+            else:
+                ax.set_title(f'{impact.capitalize()} mutations\nNot enough data for correlation')
+
+            ax.set_xlabel('dN/dS (Omega)')
+            ax.set_ylabel('dN/dS (dNdScv)')
+            ax.legend(bbox_to_anchor=(1.05, 1), loc='upper left')
+            ax.legend().remove()
```
m-huertasp: From what I understand, a legend is generated by line 71 (because you set a `hue` and you don't add `legend=False`):
```
sns.scatterplot(data=impact_subset, x='dnds_omega', y='dnds_dndscv', hue='gene', ax=ax)
```
Then you try to reposition the legend to end up removing it.

If you don't want a legend, the best thing would be to add the `legend=False` at line 71, no? 
It is not really important, just to know what you wanted 😄

bin/plot_explore_variability.py
```diff
@@ -252,6 +252,7 @@ def main(outdir, panel_regions, samples_json, all_groups_json, mutdensities, adj
     except Exception as e:
         print("Error in the process", e)
 
+    # pca plot for mutdensity and mutdensity adjusted
```
m-huertasp: Is there going to be a plot here??

subworkflows/local/plotting_qc/main.nf
```diff
@@ -49,12 +53,24 @@ workflow PLOTTING_QC {
...
+    
+    PLOTOMEGAVSOMEGAGLOBAL(all_omegas, all_omegas_globalloc, APPLYOMEGAQC.out.flagged_synonymous_cases)
 
     emit:
     mutdensity_plots        = PLOTMUTDENSITYQC.out.plots
     metrics_vs_depth_plots  = PLOTMETRICSVSDEPTHQC.out.plots
     metrics_vs_depth_tables = PLOTMETRICSVSDEPTHQC.out.tables
     flagged_omegas          = APPLYOMEGAQC.out.all_omegas_annotated
+    omega_vs_dndscv_qc      = omega_vs_dndscv_plots
+    omega_vs_omegaglobal    = PLOTOMEGAVSOMEGAGLOBAL.out.plots
```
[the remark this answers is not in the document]
m-huertasp: I've tested this, it is true. If `params.omega` is `true` but `params.omega_globalloc` is `false`, it trues to run this part and fails.

The error says that the "gene" column is missing, but the problem is that the placeholder generated when `params.globalloc` is `false` is empty, it does not have any column.

bin/omega_vs_dndscv_qc.py
efigb: _I understand that the idea of this line is to get the name of the sample groups that you'll use later to generate the plot, right?_
yes! 

 _Is the idea to plot only groups in which there is at least one flagged gene or you want to plot them all?_
Clarification here: the ideas is to discard those genes that are flagged per group to only plot those not flagged

bin/plot_explore_variability.py
efigb: this can be ommited since this was for another branch

bin/omega_vs_dndscv_qc.py
m-huertasp: I see.. What I was trying to say is that with the current strategy, from what I understand, you are only going to plot `sample_groups` in which there's at least one flagged gene. Because you are getting the `sample_groups` from the `flagged_genes_df`.
Groups in which there's not a flagged gene are not going to appear in the plot. 

Maybe it is clearer with an exampe, imagine you have this table:

| group | gene | omega | flagged |
| --- | --- | --- | --- |
| A | TP53 | 2 | No |
| B | KRAS | 1 | Yes |
| B | TERT | 2.5 | No |

In this case, you would want to plot TP53 for group A and TERT for group B. But, you are using `flagged_genes_df` to extract the sample group names that you are going to plot, which includes only:

| group | gene | omega | flagged |
| --- | --- | --- | --- |
| B | KRAS | 1 | Yes |

Then, the only group that you are going to get is B. Therefore, you are going to plot only TERT for group B,

Is this clearer?

efigb: Exactly, now I see your point, it's totally true that if any group is not flagged (like A in the example) with this strategy we will miss that group for plotting. Since we are doing the 'all_samples' comparison the gene comparison will be done for that gene, but you're right that we would miss the group comparison for that particular gene. The solution would be to extract the group information from a predefined variable for groups in deepCSA (not sure if it exists) or from the features table, but not sure which is the best way tbh...

